PR #3, opened 2016-03-19
feat(app): base structure

ami:
non-variable colorz are FPO.


app/components/dex.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from 'angular2/core';
+import { Title }     from 'angular2/platform/browser';
```
robinjoseph08: you can remove this

robinjoseph08:
LGTM

PR #32, opened 2016-03-22
feat(login): login base

ami:
（*’∀’人）♥

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/14

### DIFF
--- a/app/views/login.html
+++ b/app/views/login.html
@@ -1,1 +1,17 @@
-<h1>Hi this is login page</h1>
+<form>
+  <div class="form-alert">
+    username not found
+  </div>
+  <div class="form-group">
+    <label for="username">Username</label>
+    <input id="username" type="text" required placeholder="ashketchum10">
+    <i class="fa fa-asterisk"></i>
+  </div>
+  <div class="form-group">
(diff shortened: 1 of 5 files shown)


app/views/login.html
```diff
@@ -1 +1,17 @@
-<h1>Hi this is login page</h1>
+<form>
+  <div class="form-alert">
+    username not found
+  </div>
+  <div class="form-group">
+    <label for="username">Username</label>
+    <input id="username" type="text" required placeholder="ashketchum10">
```
robinjoseph08: are you able to limit the chars to 20 like reddit does for login and register? and limit password to 72 as well?

robinjoseph08:
one thing but lmgtlmgtlmgtlmgt :+1:

app/views/login.html
ami: should i add it for the register form too?

robinjoseph08: yup!

ami: actually it's kinda confusing to have a maxlength on creation b/c you might not be looking see that it didn't fit. nvm.

ami:
updatedddd

Issue #30, opened 2016-03-22
Finalize routing


robinjoseph08:
### why

instead of having `/tracker`, the home route would redirect you to your own dex to see, allowing anybody to go to any user's dex by going to the correct link

### notes

- change tracker route to be dynamic and pull from the username in the link
- if logged in, redirect from login, register, and home
- pull nav out of the app component and put it in each child component (login, register, tracker) since the home page won't have it

PR #39, opened 2016-03-23
feat(register): hook up the register page

### LINKED ISSUE #24: Hook up register page
### notes

- add account service
- catch errors
- have some sort of "flash" to display errors
- save returned token in a cookie or local storage

### LINKED ISSUE #29: Signout route
### notes

- have a route that just clears the cookie since no interaction with the API is needed

robinjoseph08:
ʕʘ̅͜ʘ̅ʔ

### notes

keep in might this is only for registration, nothing else, which means that even if the registration goes through successfully, you won't actually see any signs that you're logged in. to make sure it worked, you can open up the inspector, head to the Resources tab, expand the Local Storage section in the left sidebar, click http://localhost:8080 and you should see `token` and then a string that starts with `eyJ0eX`:

screen shot 2016-03-22 at 10 43 59 pm

if you want to log out, just select the `token`, delete it, and refresh.

also, since we have no staging env, lemme know after you're done testing so I can purge the db.

### things to test

- error messages
- password confirm
- optional friend code
- redirect on success

fixes #24 
fixes #29

### DIFF
--- a/app/boot.ts
+++ b/app/boot.ts
@@ -7,9 +7,11 @@ import { bootstrap }        from 'angular2/bootstrap';
 import { HTTP_PROVIDERS }   from 'angular2/http';
 import { ROUTER_PROVIDERS } from 'angular2/router';
 
+import { ApiService }   from './services/api';
 import { AppComponent } from './components/app';
 
 bootstrap(AppComponent, [
+  ApiService,
   HTTP_PROVIDERS,
   ROUTER_PROVIDERS
 ]);
--- a/app/classes/session.ts
+++ b/app/classes/session.ts
@@ -1,0 +1,9 @@
+export class Session {
+
+  public token: string;
+
+  constructor (params) {
+    this.token = params.token;
+  }
+
+}
--- a/app/classes/user.ts
+++ b/app/classes/user.ts
@@ -1,0 +1,35 @@
+const FRIEND_CODE_REGEX = /(\d{4})(?=\d)/g;
+
+export class User {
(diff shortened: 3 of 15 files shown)


ami:
...this is very hard to test roro

robinjoseph08:
yeah... okay gimme a sec

ami:
https://api.pokedextracker.com/users <- is it ok that this is public?

ami:
THE FC INPUT THINGY IS SO KEWL

robinjoseph08:
1. I think so? I mean, every user's page is open to begin with, right? soooo... but I can remove the endpoint if you think it's weird
2. I KNOW RIGHT
3. You should pull! You can see things relating to your registration now! And sign out works too

ami:
1. yea i guess. it's not a huge deal

q: do you want to do the "routing" now for register now? like that someone logged in shouldn't be able to go to /register? or is that a separate task

robinjoseph08:
if they are logged in and they go to /register, where should it redirect to? their dex page?

ami:
yeah. pretty much the only page you should be able to go to when you're signed in is the dex page (not home, login, or register), right? but you can see your own (with edit rights) or other users' (view only)

robinjoseph08:
right. then I'll do that separately when i redirect home to the dex page as well. that requires changing the route to be dynamic and what not. I've updated #30 to reflect that so I don't forget

ami:
kewl LGTM!!

Issue #43, opened 2016-03-24
404 component design


robinjoseph08:
### why

if someone goes to a non-existent page, they should should see a 404, but the url should stay the same so it should be made into a component to be reused

ami:
missingno? ;)

robinjoseph08:
yassssss

ami:
or luxio (national dex 404)??? http://bulbapedia.bulbagarden.net/wiki/Luxio_(Pok%C3%A9mon)

robinjoseph08:
oooo thats also a good one! i think im still leaning toward missingno, but i do like the idea of luxio so you can pick

PR #44, opened 2016-03-24
feat(routes): finalize routing

### LINKED ISSUE #30: Finalize routing

### LINKED ISSUE #40: Restructure app to remove nav from home
slack for ios upload

robinjoseph08:
ヘ(￣ω￣ヘ)

### what
* moved nav down to login, register, and tracker components, removing it from home
* if logged in, redirect from home, login, and register to your tracker page
* change from `/tracker` -> `/u/:username` and load correct basic info (no captures yet) of the user

### notes
* this does not handle user 404's. that will be handled in #43
* by moving nav, it messed up the flexbox configuration, so that needs to be fixed. i tried fixing it and other things broke so i'll leave that to you :)
* when a tracker page is loading, i'm hiding the entire dex component (to simulate loading and because it'll err otherwise), causing some weird stuff (the info component starts at the top). when you add the loading gif, that'll be the time to fix that
* the way it's set up right now, it would be possible to have multiple loading states (show the header after the user is done loading, then show the pokemon, then show the captures, etc). i don't know if that's desired or if you just wanna have one "loading state" that ends when everything is done loading. what do you think?

fixes #30 
fixes #40

### DIFF
--- a/app/components/app.ts
+++ b/app/components/app.ts
@@ -3,31 +3,18 @@ import { RouteConfig, RouterOutlet } from 'angular2/router';
 
 import { HomeComponent }     from './home';
 import { LoginComponent }    from './login';
-import { NavComponent }      from './nav';
 import { RegisterComponent } from './register';
-import { SessionService }    from '../services/session';
 import { TrackerComponent }  from './tracker';
 
-const HTML = require('../views/app.html');
-
 @Component({
-  directives: [NavComponent, RouterOutlet],
-  providers: [SessionService],
+  directives: [RouterOutlet],
   selector: 'app',
-  template: HTML
+  template: '<router-outlet></router-outlet>'
 })
 @RouteConfig([
   { component: HomeComponent,     name: 'Home',     path: '/' },
   { component: LoginComponent,    name: 'Login',    path: '/login' },
   { component: RegisterComponent, name: 'Register', path: '/register' },
-  { component: TrackerComponent,  name: 'Tracker',  path: '/tracker' }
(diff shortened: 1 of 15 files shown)


ami:
1 Q: should signing out route you back to /home? right now it shows you the "view only" version of your own tracker, which is slightly confusing.

re: gifs, i'd prefer to just have one gif that loads when everything's done. no point in using any one part of the tracker without any of the others.

robinjoseph08:
technically, it just leaves you on the page you were on (regardless of whether it's your tracker or not), but i can see it maybe being better to redirect home. i dont have a preference.

cool, then we'll do one gif :+1:

ami:
ic. it might be less confusing once we actually have a separate design for "viewing" someone's dex, and maybe it'll be more clear. we'll see. begs the question if we should have some sort of indication of "viewing" vs. "editing". cause you could also be signed in but be viewing someone else's dex..

anyways lgtm!

robinjoseph08:
hmm good point. yeah, we might need that distinction.

do you wanna wait to merge this until you can fix the styles or do you wanna do that in a separate PR?

ami:
Let's merge this one and I'll rebase my homepage PR and fix it there.

Issue #50, opened 2016-03-25
Design "View Only" mode of tracker page


ami:
- can not toggle captures
- no info box
- no cursor: pointer

extra note: add in percentage to both views

Issue #20, opened 2016-03-22
Rethink hover to see info UX


ami:
### why
- if i'm in the middle of a cell and i want to scroll on the info bar, it's essentially impossible

### thoughts
- get rid of hover? add another click instead?

ami:
@robinjoseph08 a couple ideas i have below. lemme know what you think!

![slack for ios upload 1](https://cloud.githubusercontent.com/assets/5498072/14009373/770ef766-f148-11e5-88f0-fed7643db649.jpg)

1) **split on hover:** on hover, split the panel in half. one half is the button to show `info`, one half is the button to mark as `captured`

**pros**:
- we get to keep the existing look of the box view
- it's very clear on hover the two things you can do

**cons**:
- what do we do AFTER a button is clicked? animation for marking as `captured`?

2) **info btn:** button across the bottom of each panel to open `info`, replacing the space that `national_id` currently takes up. the rest of the box remains as the way to mark as `captured`

**pros**:
- no hover action, so it's a simpler experience

**cons**:
- will so many "info" buttons on the main box view look messy?
- we lose information about the pokemon (`national_id`)

3) **info btn on hover:** show a button on hover in the corner of the panel to open `info`. the rest of the box remains as the way to mark as `captured`

**pros**:
- we get to keep the existing look of the box view
- still lots of room to mark as captured

**cons**:
- does this draw too much attention to the "info" button and not enough to the action of marking as `captured`? perhaps we'd need to add an additional hover style to the rest of the box to show that that part of the box is clickable too.

robinjoseph08:
these are some good ideas :+1: i dont know which one i prefer just yet, but ive got a few comments for them:

- will the info panel still default to bulbasaur for all of these?
- for 1, yeah it's a bit weird to think about what happens on click... maybe the background of the captured side turns yellow and then when you hover off of it, you see the whole box is yellow? i don't think that's ideal
- at first glance i really like 2, but i think the number is actually pretty important info (since everything is organized by this number), so it shouldn't be removed. also a whole 30-pokemon box of these with little "info" bars might break the nice clean look we've got going on and i like that nice clean look
- i'm concerned for how difficult it would be to click the info button for 2 and 3. is it too small? i liked how easy it was to show the info with our initial hover idea, so ideally, it stays as easy to show the info. i think it might be fine, but it's just something i'm slightly worried about
- i would prefer that the image always be visible (so that's a point in favor of 2 and 3)
[...]

PR #51, opened 2016-03-25
feat(captures): load, create, and delete captures

### LINKED ISSUE #25: Load, save, and delete captures
### notes

- ~~dependent on #30~~ ready to start working on it
- add captures service (`list`, `create`, `delete`)
- add `captured` class if `capture.captured === true`
- hook up the progress bar/numbers (only for national)
- add the ability to toggle the captured state if it's your own dex

robinjoseph08:
(๑ॢ˃̶͈̀ ꇴ ˂̶͈́๑ॢ)

### notes

- i think we need some sort of "pokemon loading" state when the request to toggle captured state is going through. cause if you click it twice fast, it'll send two `captured = true` requests and even though nothing bad happens (since the second one just fails), ideally, it doesn't make the request.
- also something that happens when you click fast, it sometimes selects the icon/name/number, so maybe disable selection for the pokemon box? idk if people want to be able to copy and paste the pokemon names though...
- for when you aren't logged in (or you're viewing someone else's dex), ive disabled the ability to toggle captures, but the `cursor: pointer` style is still on it, so we should have a way that turns it back to `default`.
- in the same vein, for the empty pokemon blocks in the last box (721-750), it also has `cursor: pointer` so `empty` class should be `default` as well
- the on-hover-the-info-panel-changes flow might be a little sluggish. i decided to not worry about that for now considering the fact that we're gonna be removing the hover event anyway. just keep that in mind.
- right now, it still takes a decent amount of time to load the page but i _think_ it's because the bundle.js (that's generated on the fly and is a whopping **14.3MB**) is slow. that'll be fixed once it's deployed and the bundle has been minified/uglified/gzipped, just like what we do in dashboard.
- the `/captures` endpoint is still a tad slow for my taste, but im thinking of ways to make it faster. one major limiting factor is the fact that we're "generating" fake captures if they don't exist in the DB. that iteration (check if a capture exists, if not make one) is considerably slower than a regular DB fetch (~120ms vs ~5ms), so that'd be the place to improve. idk, we'll see.

fixes #25

### DIFF
--- a/app/classes/capture.ts
+++ b/app/classes/capture.ts
@@ -1,0 +1,15 @@
+import { Pokemon } from './pokemon';
+
+export class Capture {
+
+  public user_id: number;
+  public pokemon: Pokemon;
+  public captured: boolean;
+
+  constructor (params) {
+    this.user_id = params.user_id;
+    this.pokemon = new Pokemon(params.pokemon);
+    this.captured = params.captured;
+  }
+
+}
--- a/app/components/box.ts
+++ b/app/components/box.ts
@@ -1,7 +1,7 @@
 import { Component, EventEmitter } from 'angular2/core';
 
+import { Capture }          from '../classes/capture';
 import { NumberPipe }       from '../pipes/number';
-import { Pokemon }          from '../classes/pokemon';
 import { PokemonComponent } from './pokemon';
 
 const HTML = require('../views/box.html');
@@ -9,21 +9,21 @@ const HTML = require('../views/box.html');
 @Component({
   directives: [PokemonComponent],
   events: ['pokemonHover'],
-  inputs: ['pokemon'],
+  inputs: ['captures'],
   pipes: [NumberPipe],
   selector: 'box',
   template: HTML
 })
 export class BoxComponent {
 
-  public pokemon: Pokemon[];
+  public captures: Capture[];
 
-  public pokemonHover = new EventEmitter<Pokemon>();
+  public pokemonHover = new EventEmitter<Capture>();
 
   private boxSize = 30;
 
   public get empties (): Object[] {
-    return new Array(this.boxSize - this.pokemon.length);
+    return new Array(this.boxSize - this.captures.length);
   }
 
 }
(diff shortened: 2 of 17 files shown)


ami:
bahhhhhhhhhhhhhhhhhhhhhhhhhhh i don't wanna rebaseee

robinjoseph08:
lol if you can get yours in soon, ill deal with the rebase

ami:
it's ok mine's not ready yet anyways haha. LGTM!! 

- the requests are happening pretty fast for me, so i'm not sure if this is necessary? in reality, no person is going to be tapping a single box fast multiple times (only we would do it during testing) so i don't think it's necessary, especially since nothing bad happens on the API side.
- i'd rather not disable it in case you do want to select and copy something in the box. plus, like i mentioned above, people really shouldn't normally be clicking on them quickly like that anyway.
- :+1: added that to https://github.com/robinjoseph08/pokedextracker.com/issues/50
- :+1: will make sure i do it in https://github.com/robinjoseph08/pokedextracker.com/issues/20
- :+1: to the rest

ami:
gonna merge this so i can rebase my branch. but feel free to continue responding ^

robinjoseph08:
* I'd still like to consider the loading state, not really to disable cause I agree that most people won't be clicking it fast, but more for feedback for the user. Cause while it has been quite snappy, I don't think it'll always be that way (a user's internet connection is slow, our tiny server is under load, etc). I was messing with it on mobile with spotty Internet and it felt laggy cause when I tapped, nothing happened (since the request was being made). Another possibility other than a loading state is to actually toggle the capture state _before_ making the request to the server as opposed to after. That way, the user sees immediate feedback, the request can take however long it wants, and _if_ there happens to be a failure from the server side (which there shouldn't really ever be), we can just revert it back and maybe display a hovering error flash or something something. I'd have to make some API changes to accommodate this (essentially making it idempotent, so that if a request comes in to mark a Pokemon as captured and it's already captured, it doesn't error, it just lets the request through), but that wouldn't be too hard. What do you think is the better experience? It might be weird to show that the Pokemon is captured even though we haven't saved it just yet, I don't think there are any downsides. 
* Cool, I can dig it :+1: yeah people won't really be clicking rapidly. 
* :+1: to the rest

ami:
- okie doke. it could be though because of the transition animation on .pokemon-toggle (which I've removed in my branch). i think the loading state idea is better than changing the API, if we do go with one

PR #56, opened 2016-03-29
fix(dex): replace hover ux with clicks

ami:
(:ᘌꇤ⁐ꃳ 三

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/20


robinjoseph08:
added my commit, but it's pretty much just a rename soooo...

LGTM!

PR #62, opened 2016-03-30
feat(favicon): add favicon

ami:
☜(⌒▽⌒)☞

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/31


robinjoseph08:
idk if im crazy, but the images seems a bit fuzzy to me? i think it might be that the anti-aliasing is a bit excessive. from a cursory google search of "[pixel art anti aliasing](http://lmgtfy.com/?q=pixel+art+anti+aliasing)" shows that you need to do the AA custom? idk, that seems like a lot of effort lol

do you also think it's a bit blurry?

ami:
it could be b/c your screen is retina and we're just displaying the images at full size - not half size like you're supposed to with retina?

ami:
wait nvm I resized you were talking about the favicon, not the bulbapedia images

you're right, it is slightly blurry - reason being that the design I did has a "center" column, so the images need to be a odd number of pixels wide/high - but the favicons are even let sized(16x16, 32x32, etc). maybe later I can go in and fix it to not have a "center" but at the time I didn't want to lol ;_____;

robinjoseph08:
ahh yeah i think that might be it.

...wanna do it now?（＾_＾）pretty pls? a crisp favicon is a happy favicon!

ami:
finnneeee updated

robinjoseph08:
looks like you need a rebase :see_no_evil:

ami:
done! and made a new favicon, check it out.

ami:
YAYYY

ami:
(updated)

robinjoseph08:
:heart_eyes: perfecto :ok_hand: LGTM!!

PR #65, opened 2016-04-01
fix(font): use font awesome male and female signs

### LINKED ISSUE #12: Replace male/female symbol font
### why
- the font we're using for display doesn't have those symbols, so it's defaulting to helvetica which looks very ugly (for nidoran m/f)

### notes
- find a way to inject & replace with fontawesome (`fa-mars` & `fa-venus`) so that the info stays as-is in the database

robinjoseph08:
ʕ ⊃･ ◡ ･ ʔ⊃

this still needs your styles cause right now, they're pretty darn huge

fixes #12

### DIFF
--- a/app/classes/pokemon.ts
+++ b/app/classes/pokemon.ts
@@ -20,7 +20,7 @@ export class Pokemon {
 
   constructor (params) {
     this.national_id = params.national_id;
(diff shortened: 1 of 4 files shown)


ami:
styles added. lgtm!

Issue #17, opened 2016-03-22
Collapsable info panel


ami:
### why
- it takes up a lot of room, let people who don't want it hide it.

robinjoseph08:
Do you think we should save the preference of whether they like it closed or open locally so that even if they refresh, it'll stay how they want?

ami:
👍

PR #66, opened 2016-04-01
feat(info): collapsible info panel

ami:
꒳ᵃ꒳ᵃ꒳ᵃ~(๑°ᗨૢ°๑)♡ӵᵉ੨ᑋ✧

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/17

#### what
- adds collapsible info panel
- uses that collapse functionality for a better responsive/mobile experience

#### notes
still needs:
- [x] detect browser width on load, and collapse info panel if width < 1100px (but not if someone resizes their window to below the threshold, in that case i think it should stay open)
- [x] if the panel is collapsed, when someone clicks the info button on a pokemon, it should open

### DIFF
--- a/app/components/tracker.ts
+++ b/app/components/tracker.ts
@@ -25,6 +25,7 @@ export class TrackerComponent implements OnInit {
   public active: Capture;
   public captures: Capture[] = [];
   public loading: boolean = true;
+  public collapsed: boolean = false;
   public _session: SessionService;
   public user: User;
 
--- a/app/components/tracker.ts
+++ b/app/components/tracker.ts
@@ -14,6 +14,8 @@ import { UserService }       from '../services/user';
 
 const HTML = require('../views/tracker.html');
 
+const MOBILE_WIDTH = 1100;
+
 @Component({
   directives: [DexComponent, InfoComponent, NavComponent, NotFoundComponent],
   providers: [CaptureService, SessionService, Title, UserService],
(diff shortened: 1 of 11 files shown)


app/views/info.html
```diff
@@ -1,10 +1,18 @@
 <div *ngIf="active" class="info-header">
   <img [src]="active.pokemon.icon_url">
-  <h1 [innerHTML]="active.pokemon.name"></h1>
+  <h1></h1>
```
robinjoseph08: is there a reason you left this empty element here? is it necessary?

robinjoseph08:
wait, am i missing something? this is what it looks like for me:

screen shot 2016-04-03 at 8 57 21 am

ami:
lol I must have messed up the rebase. oopz

ami:
@robinjoseph08 fixed! lols

robinjoseph08:
@ami at what width should it it "auto-collapse" on load? <= 400px?

also, some intermediary sizes are a bit wonky. are you gonna deal with that separately?

screen shot 2016-04-03 at 7 50 44 pm

robinjoseph08:
also the mobile info button hover effects are still there for disabled pokemon. should i create a separate issue or is that a quick fix?

app/components/tracker.ts
```diff
@@ -14,6 +14,8 @@ import { UserService }       from '../services/user';
 
 const HTML = require('../views/tracker.html');
 
+const MOBILE_WIDTH = 750;
```
ami: this should be 1100 my b

PR #79, opened 2016-04-08
feat(tracker): design view only mode

ami:
(っ- ‸ – ς)
^ sad because angular 2 is confuzzling

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/50

i need a `viewing` class on `<pokemon>` if someone is viewing in order to disable the hover effects. but i couldn't get the angular part to work lolol. can you help me add that? the css is already set up with the class `viewing`.

### DIFF
--- a/app/components/box.ts
+++ b/app/components/box.ts
@@ -3,14 +3,17 @@ import { Component, EventEmitter } from 'angular2/core';
 import { Capture }          from '../classes/capture';
 import { NumberPipe }       from '../pipes/number';
 import { PokemonComponent } from './pokemon';
+import { SessionService }   from '../services/session';
+import { User }             from '../classes/user';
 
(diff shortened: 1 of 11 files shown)


robinjoseph08:
@ami added!

I also introduced the [**elvis operator**](http://julienrenaux.fr/2015/12/02/angular2-series-template-syntax/#Elvis_operator)! It's syntactic sugar for checking if a value is `undefined` first. It essentially makes expressions as forgiving as they were back in angular 1:

```html
<div *ngIf="_session.user && _session.user.admin">
<!-- turns into -->
<div *ngIf="_session.user?.admin">
```

｡:.ﾟヽ(´∀`｡)ﾉﾟ.:｡+ﾟ
yayayayay angular 2!

ami:
lgtm!

PR #106, opened 2016-04-12
fix(tracker): loading text

ami:
¯\_༼ ಥ ‿ ಥ ༽_/¯

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/22

i can't get the loading part in `info.ts` to work lol. fix itttt

### DIFF
--- a/app/classes/capture.ts
+++ b/app/classes/capture.ts
@@ -17,8 +17,8 @@ export class Capture {
     this._pokemon = _pokemon;
   }
 
-  public loadPokemon () {
-    this._pokemon.retrieve(this.pokemon.national_id)
+  public loadPokemon (): Promise<Pokemon> {
+    return this._pokemon.retrieve(this.pokemon.national_id)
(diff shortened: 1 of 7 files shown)


robinjoseph08:
@ami it's now hooked up! ...but the entire info panel disappears while it loads lol. fix itttt

ami:
@robinjoseph08 fixeddd

robinjoseph08:
what do you think about adding `LOADING...` to the info panel too? the fact that everything disappears a bit weird to me. or maybe just make the `.info-location` disappear and leave the name, icon, and number in the header (since that part doesn't need to load)?

ami:
i tried that but it looked even weirder b/c it's only there for such amount of time. didn't realize the header doesn't need to load - i think that'd be ideal. i'll try that!

ami:
updateddd

robinjoseph08:
:boom: looks bootiful

Issue #16, opened 2016-03-22
Select by entire row and/or box at once


ami:
### why
- makes it easier to input what you already have upon signup

### notes
- this may need to get broken up, may need API and/or design & frontend things
- depends on #101

PR #107, opened 2016-04-12
feat(box): mark/unmark entire box

ami:
fixes https://github.com/robinjoseph08/pokedextracker.com/issues/16

needs to be connected to the back-end. i'm thinking that it will change from "mark all" to "unmark all" if you have the whole box already. also, while the request is happening, it'd be cool to disable the button and replace the content with a spinner of some sort?

### DIFF
--- a/app/components/box.ts
+++ b/app/components/box.ts
@@ -2,6 +2,7 @@ import { Component, EventEmitter } from 'angular2/core';
 import { DecimalPipe }             from 'angular2/common';
 
 import { Capture }          from '../classes/capture';
+import { CaptureService }   from '../services/capture';
 import { PokemonComponent } from './pokemon';
 import { SessionService }   from '../services/session';
 import { User }             from '../classes/user';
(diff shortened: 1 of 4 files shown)


robinjoseph08:
@ami updated! ⊂(・ヮ・⊂)

* add functionality to the button
* change to "unmark all" if all of the pokemon are caught
* disable the button when the mark all request is loading
* insert the font awesome spinner in the button while loading
  * this needs to be styled
  * it's a bit jittery on my computer, but everything is laggy on my computer right now... and idk how it will fare once deployed

also... the update to api i had to make kinda broke things in prod... so once this is merged, we'll either need to deploy this or revert the api until we deploy this...

ami:
@robinjoseph08 LOL wth?? i tested that it works (it's pretty fast and not laggy for me!), so you should go ahead and revert API so prod is fixed. i should be fine updating the styles without the actual implementation.

robinjoseph08:
@ami reverted! ^_^

and good to hear about the lag. time to restart my computer :+1:

ami:
@robinjoseph08 updated!!

ami:
also this was totes easy. we should do this on dashbooaorrdddd

robinjoseph08:
looks siiiick :+1: 

and fo sho. we can do it on friday!

ami:
yayyyy fridaysss

PR #114, opened 2016-04-16
feat(analytics): track all the things!

robinjoseph08:
૮( ᵒ̌ૢ௰ᵒ̌ૢ )ა

im learning all sorts of things about google analytics! did you know you can track custom events? and it turns out our page tracking was wrong since it's all a single page app (russell pointed that out to me). thats why we never saw `/login` or `/register` page views, cause it wasn't correctly tracking page transitions, only the first landing.

so this pr fixes that and adds tracking for several ga events. a ga event consists of an `action`, a `category`, an optional `label`, and an optional `value`. i use the first 3, but i don't use `value` cause it has to be an integer, and i dont like that.

the events ive added tracking for:

category | action | label | notes
---|---|---|---
Session | login
Session | register
Session | sign out
Pokemon | mark | the pokemon's name
Pokemon | unmark | the pokemon's name
Pokemon | show info | the pokemon's name
Box | mark all | the number range e.g. `001 - 030`
Box | unmark all | the number range e.g. `001 - 030`
Info | open
Info | close
Region | open | | this is for mobile
Region | close* | | this is for mobile
Region | toggle | the region they toggled to
Share | open
Share | close*
Share | select link

_*this is only triggered if they close it "correctly". if they close it by clicking outside, it doesn't trigger the event._

i can show you what the events look like on the ga dashboard to you can see how these properties map to things. if you think i missed an event, lmk!

### DIFF
--- a/app/boot.ts
+++ b/app/boot.ts
@@ -7,6 +7,7 @@ import { bootstrap }        from 'angular2/bootstrap';
 import { enableProdMode }   from 'angular2/core';
 import { HTTP_PROVIDERS }   from 'angular2/http';
 import { ROUTER_PROVIDERS } from 'angular2/router';
+import { Angulartics2 }     from 'angulartics2';
 
 import { ApiService }   from './services/api';
 import { AppComponent } from './components/app';
@@ -16,6 +17,7 @@ if (process.env.NODE_ENV === 'production') {
 }
 
 bootstrap(AppComponent, [
+  Angulartics2,
   ApiService,
   HTTP_PROVIDERS,
   ROUTER_PROVIDERS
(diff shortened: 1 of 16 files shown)


ami:
what about adding an event for clicking "view on bulbapedia" and saving the Pokemon name? that's the only remaining one I can think of!

also, does the IP whitelisting apply for these events as well (will our events be filtered out)?

robinjoseph08:
cool, added:

category | action | label
---|---|---
Info | open Bulbapedia link | the pokemon's name

and yeah, the filtering filters literally everything 👍 and I even think theres a better way we can filter too so we can view the site from anywhere, but we'll deal with that later

robinjoseph08:
changed the info events to collapse and uncollapse

ami:
lgtm!

Issue #122, opened 2016-04-19
fix info panel for firefox (flexbox)


ami:
https://i.imgsafe.org/03840c9.png

PR #123, opened 2016-04-20
fix(styles): fix firefox bugs

ami:
щ(ಠ益ಠщ)

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/122, if you are curious here is the reasoning: http://stackoverflow.com/questions/33790219/firefox-not-respecting-flex-shrink

ss of firefox below, verified that it still works in chrome and mobile. since my safari is still broken... can you verify that? interestingly enough, prod (pokedextracker.com) is totally fine for me in safari, it's only broken locally. CLUES!!!

screenshot 2016-04-19 21 39 23

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -29,6 +29,8 @@ app {
     @include flex(1 1 auto);
     @include flexbox();
     @include flex-direction(column);
+
+    min-height: 0;
   }
 }
 
@@ -136,6 +138,10 @@ input {
 
   &:focus {
     border-color: $brand-secondary-dark;
+  }
+
+  &:invalid {
+    box-shadow: none;
   }
 }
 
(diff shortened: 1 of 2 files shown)


ami:
nvm my local safari works, it's just ridiculously slow o.o

robinjoseph08:
kewl, thanks for the SO link 👍 

also another thing, that idk if its _super_ relevant, if you have a small height and looking at eevee:

screen shot 2016-04-20 at 9 48 36 pm

weird stuff starts going down. but it seems to be happening in chrome and on prod too, so ¯\\\_(ツ)_/¯

other than that, FF LGTM!

ship it

ami:
yeaaaaa meh. lol. i'll figure that out later

Issue #117, opened 2016-04-17
back to top button/scrollbar


robinjoseph08:

robinjoseph08:
possibly having a scrollbar visible as well would be nice

robinjoseph08:
it seems like the scrollbar is more desired than the back to top button. we can mess with the scrollbar pseudo elements `dex::-webkit-scrollbar` but idk how well it works with the info panel and whatnot: https://css-tricks.com/custom-scrollbars-in-webkit/ i tried messing with it a little and then realized you should probably be doing this instead lol

PR #129, opened 2016-04-24
feat(dex): clickable scroll bar + back to top btn

ami:
ෆු(*˃ர்˂*)ෆු

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/117

can you add the actual back-to-top functionality for the button? also, ideally it's not there if you're at the top and only shows up maybe 50-100px down. also it'd be cool if it animated and slid in from the bottom (bonus cookie points!!).

one caveat for the scrollbar now being within the dex area + clickable - if you're in a weird "mid" state, but the info panel is expanded, the scroll bar is on the right within the info area. it's pretty unavoidable though since, in that mode, the panel is essentially absolutely positioned. i think it's fine though, because the rest of the time the scrollbar is exposed as expected.

screenshot 2016-04-23 21 57 35

### DIFF
--- a/app/components/tracker.ts
+++ b/app/components/tracker.ts
@@ -15,7 +15,8 @@ import { UserService }       from '../services/user';
 
 const HTML = require('../views/tracker.html');
 
-const MOBILE_WIDTH = 1100;
+const MOBILE_WIDTH          = 1100;
+const SHOW_SCROLL_THRESHOLD = 400;
 
 @Component({
   directives: [DexComponent, InfoComponent, NavComponent, NotFoundComponent],
@@ -30,6 +31,7 @@ export class TrackerComponent implements OnInit {
   public loading: boolean = true;
   public collapsed: boolean = false;
   public _session: SessionService;
+  public showScroll: boolean = false;
   public user: User;
 
   private _capture: CaptureService;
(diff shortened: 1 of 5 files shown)


robinjoseph08:
@ami add the functionality! and hopefully the sliding in is worthy of the 🍪 points! but 50-100 seemed instantaneous for me, so i made it 900px instead lol. if it's too much we can change it.

i touched your css (๑꒪̇⌄꒪̇๑) i hope i did it in a way that's at least somewhat acceptable. but if not, feel free to fix it up lol

because of performance issues locally, i deployment this branch to staging and i _think_ it's okay. you can be the final judge though. i tried doing some angular animation stuff to accomplish it, and it was absolutely terrible. so we definitely gotta go css. but if this isn't good, we can see what other options there are.

as mentioned before, right now, the scroll to top just jumps to the top, but we can look into animating that separately.

app/components/tracker.ts
```diff
@@ -15,7 +15,8 @@ import { UserService }       from '../services/user';
 
 const HTML = require('../views/tracker.html');
 
-const MOBILE_WIDTH = 1100;
+const MOBILE_WIDTH          = 1100;
+const SHOW_SCROLL_THRESHOLD = 900;
```
ami: i think 900 is a bit much (but you're right, 50/100 was def too little). i think it should show up around when you've just hit the top of the first box - how's 400?

app/styles/tracker.scss
```diff
@@ -262,6 +263,13 @@ dex {
       cursor: pointer;
       background-color: $brand-secondary-dark;
     }
+
+    &.visible {
+      @include transition(all .5s ease-in);
```
ami: i don't think you need this `@include` - just the one under `.scroll-up` is fine

robinjoseph08:
@ami updated!

ami:
lgtmmm

PR #131, opened 2016-05-01
feat(config): add configs for each environment

robinjoseph08:
(๑•̀ㅂ•́)و✧

**finally** adds the configs to support a staging env!

you should run it and make sure it still runs, but now, itll be talking to the staging api. so no accounts exist there. i made another `ashketchum10` so the homepage works and i populated the db with pokemans so it should be good to go!


ami:
lgtmmmm

PR #134, opened 2016-05-18
feat(register): send referrer

robinjoseph08:
┻┳|･ω･)ﾉ

adds saving the referrer on signup. here's a redirect link to [localhost:8080](http://goo.gl/fT3cc4) so that it sets the value of `document.referrer` when youre running locally (you should open console and type `document.referrer` to make sure it's set). then register. then i can either give you staging db access or i can verify it worked for you.


ami:
.... i believe you 😄

PR #135, opened 2016-05-22
feat(evolutions): add links to the pokemon

### LINKED ISSUE #95: Make pokemon in evolution families clickable (change info panel)

robinjoseph08:
(੭ु˙꒳​˙)੭ु⁾⁾

fixes #95 

idk if i was supposed to add the `cursor: pointer` or just add an empty `href` ¯\\\_(ツ)_/¯ if the `href` is preferred, i can change it.

also, idk if we want an active state or something. 

another thing to note is the fact that i can still click on a selected pokemon and it will "reload". should we disable that or is it not a big deal? i dont really see any harm in it. what do you think?

### DIFF
--- a/app/classes/evolution-family.ts
+++ b/app/classes/evolution-family.ts
@@ -1,13 +1,14 @@
 import { Evolution } from './evolution';
+import { Pokemon }   from './pokemon';
 
 export class EvolutionFamily {
 
-  public pokemon;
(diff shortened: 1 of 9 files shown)


ami:
lgtm! i don't think either of those is necessary (adding an active state or disabling the pokemon you're already on). maybe a hover state, but since there's the cursor, i think it's good enough.

PR #149, opened 2016-09-05
chore(deps): upgrade angular beta.17 -> rc.4

robinjoseph08:
ƪ(΄◞ิ۝◟ิ‵)ʃ

upgraaaaade! this one was quite a doozy. this currently isnt the latest rc (`rc.6` came out 8/31) but there was a huuuge change in structure from `rc.4` to `rc.5` so i thought id take things slow

changes include (for documentation purposes):

- getting rid of the `angular2` module in favor of `@angular/*`
- add `@angular2/forms@0.2.0` to move to the new way forms are being handled
  - add `disableDeprecatedForms` and `provideForms` to `boot.ts`
  - add `name` attributes to the any `input` element we had `ngModel` on
- `ngSwitchWhen` ➡️ `ngSwitchCase`
- upgrade `angulartics2` from `v1.0.10` to `v1.1.5` **though i still dont think page view tracking works, only event tracking. i think when we stop using the deprecated router, it will work again, but thatll be coming in another pr**
- pass around an html sanitizer so that we dont get a bunch of dumb warnings in the console. this was caused because we are using `[innerHTML]`. i _think_ we can actually get rid of all this stuff when we upgrade to `rc.5` because of [this commit](https://github.com/angular/angular/commit/482c019) but well see
- upgrade the polyfill url from `beta.12` to `beta.17`. i forgot to do this in the last upgrade pr but it looks like there arent hosted polyfill files for the `rc`s. i think we can use [this starter repo](https://github.com/angular/angular2-seed) to correctly load the polyfills/dependencies without the need of `angular2-polyfills.js`

before moving on to the other `rc`s, i wanna move to the new router cause i think the old deprecated one doesnt work with all the new stuff

@ami same as before, all you need to do is checkout the branch, `rm -rf node_modules`, `npm i`, `npm start`, and see if all of the pages/functionality still works!

### DIFF
--- a/app/boot.ts
+++ b/app/boot.ts
@@ -3,11 +3,12 @@ import 'reflect-metadata';
 import 'rxjs';
 import './styles';
 
-import { bootstrap }        from 'angular2/platform/browser';
-import { enableProdMode }   from 'angular2/core';
-import { HTTP_PROVIDERS }   from 'angular2/http';
-import { ROUTER_PROVIDERS } from 'angular2/router';
-import { Angulartics2 }     from 'angulartics2';
+import { bootstrap }                            from '@angular/platform-browser-dynamic';
+import { disableDeprecatedForms, provideForms } from '@angular/forms';
+import { enableProdMode }                       from '@angular/core';
+import { HTTP_PROVIDERS }                       from '@angular/http';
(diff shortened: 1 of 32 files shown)


ami:
LGTM

PR #150, opened 2016-09-05
feat(twitter): add twitter link

robinjoseph08:
⋛⋋( ‘◇’)⋌⋚

add tweeter

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -73,6 +73,10 @@ home {
         }
       }
     }
+
+    .social {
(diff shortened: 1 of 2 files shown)


app/styles/home.scss
```diff
@@ -73,6 +73,10 @@ home {
         }
       }
     }
+
+    .social {
+      margin: 30px 0;
```
ami: `margin-top: 30px` instead

ami:
lgtm after change

PR #152, opened 2016-09-05
chore(deps): remove test dependencies

robinjoseph08:
（￣□￣；）

remove test modules since... we dont have tests yet. when we decide to add them, well add them back at the latest versions. all theyre doing now is getting outdated


ami:
lgtm

PR #157, opened 2016-09-06
feat(sprites): use modified pokesprite

### LINKED ISSUE #52: Think about hosting our own images
### notes

since we're just pulling assets from bulba's cdn (which is through http), it's making the "encrypted" thingy on chrome not green anymore:

screen shot 2016-03-25 at 10 20 24 pm

vs.

screen shot 2016-03-25 at 10 20 42 pm

i know there's no pressing need to host our assets, but it wouldn't be too terribly difficult to do either. just something to think about.

### LINKED ISSUE #126: Switch from using bulbapedia icon urls to pokesprite
https://github.com/msikma/pokesprite

* it might be a bit more efficient
* since it won't be pulling it through http from bulba, it will effectively close #52 
* it'll pave the way for shiny tracking
* seemed to be easier to maintain/edit/fine-tune than the scraping method

ami:
(ﾉ>ω<)ﾉ :｡･:*:･ﾟ’★,｡･:*:･ﾟ’☆

WOOOOOOoooooooooOOOOOOoooOOOOOO!!!

fixes #126 
fixes #52


robinjoseph08:
ooooooooOOOOOOOooooooo so criiisp! lgtm!!

PR #160, opened 2016-09-07
fix(spinner): hide content when spinning

ami:
(๑ˇ ῁̫ ˇ)˒˒
^ smug 🐛  catching face

the disabled spinner button on the account page wasn't actually working as expected b/c all the styles for it were nested under `box .box-header button`.


robinjoseph08:
csscsscsscsscsscss lgtm!

PR #162, opened 2016-09-09
fix(home): replace loveball w pokesprite

ami:
♡〜٩(^▿^)۶〜♡

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/159

added some manual styles since this is the only one of the "non-pokemon" sprites we're using

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -92,7 +92,10 @@ home {
     color: $brand-secondary;
     z-index: 1;
 
-    img {
+    i {
+      width: 30px;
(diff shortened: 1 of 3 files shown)


robinjoseph08:
lgtm!

not sure if it was intentional or not, but the spacing changed a bit with this pokesprite:

old | new
---|---
screenshot 2016-09-11 15 16 42 | screenshot 2016-09-11 15 16 30

ami:
yea, that's just due to sticking with the sizing of pokesprite's images

Issue #164, opened 2016-09-10
Error on safari preventing basic functionality


robinjoseph08:
when running pokedextracker.com on safari, there's an error that sometimes occurs that prevents all subsequent things from working (marking pokemon as captured, opening the info panel, showing error messages when logging in, etc). here's the error (though it's not super helpful in its current form):

<img width="727" alt="screenshot 2016-09-10 09 04 55" src="https://cloud.githubusercontent.com/assets/2454505/18411872/391fdda2-7736-11e6-9a0c-026fd7442269.png">

weird things:

- it only happens sometimes. i refreshed 3 times when logged in and on the tracker page and the first 2 times, it was fine but then it erred on the third
- i can't seem to reproduce it on chrome or firefox (though since it's non-deterministic, im not entirely sure its exclusive to safari)

ill probably have to dig into the bundle itself to see what the issue is. it might have to do with the minifying/uglifying we're doing only in prod. dont really know as of now

ami:
@robinjoseph08 updated the description since we have now seen the issue in both prod and staging.

PR #170, opened 2016-09-15
chore(deps): downgrade from angular-rc.4 -> angular-beta.17

### LINKED ISSUE #166: The site is laggy
the main suspects we think this can be from is pokesprite and the angular upgrade because those are the biggest structural changes we made. since pokesprite is pretty much just css though, im doubtful its that. that leaves the angular upgrade to blame.

im going to attempt to manually roll the upgrade back and see if that makes it better cause this seems to be affecting several people.

robinjoseph08:
( ఠൠఠ )

fixes #166 

doooowngrade. just do the usual (rm rf and npm i) and then make sure everything still works (specifically pokesprite and the account page). well deploy this guy and see what happens ¯\\\_(ツ)_/¯

### DIFF
--- a/app/boot.ts
+++ b/app/boot.ts
@@ -3,12 +3,11 @@ import 'reflect-metadata';
 import 'rxjs';
 import './styles';
 
-import { bootstrap }                            from '@angular/platform-browser-dynamic';
-import { disableDeprecatedForms, provideForms } from '@angular/forms';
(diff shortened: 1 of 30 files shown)


robinjoseph08:
since this wasnt necessary to fix #166 and it doesnt fix #164, i dont think we need this downgrade anymore

Issue #165, opened 2016-09-10
Safari doesn't have crisp edges for sprites


robinjoseph08:
| safari | chrome |
| --- | --- |
| <img width="111" alt="screenshot 2016-09-10 09 29 47" src="https://cloud.githubusercontent.com/assets/2454505/18411990/457684ae-7739-11e6-9d19-b00ffc6b7e95.png"> | <img width="112" alt="screenshot 2016-09-10 09 30 08" src="https://cloud.githubusercontent.com/assets/2454505/18411991/4ee6b2de-7739-11e6-82e4-2853c662f8a1.png"> |

ami:
https://jsfiddle.net/tzv24nay/

^ open in safari - looks like the property doesn't work for background images 😭

[this](https://bugs.webkit.org/show_bug.cgi?id=97991) seems relevent but i have no idea what it means haha

ami:
closing as this works in safari now!

PR #172, opened 2016-09-16
fix(sprites): remove pokesprite

ami:
。ﾟ(ﾟﾉД｀ﾟ)ﾟ｡

FIXES https://github.com/robinjoseph08/pokedextracker.com/issues/166 (and fixes https://github.com/robinjoseph08/pokedextracker.com/issues/165)

ideally we can bring back pokesprite in a way that doesn't kill the site usability b/c we really want to use it for enabling shiny sprites! created an issue [here](https://github.com/robinjoseph08/pokedextracker.com/issues/171) for that.


ami:
closing!

PR #173, opened 2016-09-18
fix(pkicon): better class property

### LINKED ISSUE #166: The site is laggy

robinjoseph08:
٩(˃̶͈̀௰˂̶͈́)و

FIXES #166 

shouldve done it this way to begin with, but this should fix it! you should test it fully across browsers though to make sure

### DIFF
--- a/app/classes/pokemon.ts
+++ b/app/classes/pokemon.ts
@@ -28,6 +28,7 @@ export class Pokemon {
   public or_locations: string[];
   public as_locations: string[];
   public evolution_family: EvolutionFamily;
+  public icon_class: Object;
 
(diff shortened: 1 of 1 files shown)


ami:
woohoo! lgtm!

safari still has the random error, sometimes it will even freak out and ask me to force quit. but yay!

PR #174, opened 2016-09-21
refactor(buttons): add .btn styles

robinjoseph08:
(๑•̌.•̑๑)ˀ̣ˀ̣
^ me confused by the css

i attempted to make the more generic `btn` class! arent ya proud? ^_^

that said, i literally have no idea why i had to add that `padding-top: 14px;`. i looked through all of the computed styles of the button on prod and couldnt figure out what styles were missing that made it look like this:

screenshot 2016-09-20 21 51 26

all i know is that `padding-top: 14px;` fixeded it. so if there's a better way, just lmk and ill do that lol. but everything else looks good though!

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -56,7 +56,7 @@ home {
       margin-bottom: 0;
     }
 
-    button {
+    .btn {
       display: inline-block;
       width: 180px;
 
@@ -125,7 +125,7 @@ home {
         margin-bottom: 15px;
       }
 
-      button {
+      .btn {
         width: 130px;
       }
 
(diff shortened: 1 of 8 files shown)


app/styles/home.scss
```diff
@@ -56,9 +56,10 @@ home {
       margin-bottom: 0;
     }
 
-    button {
+    .btn {
       display: inline-block;
       width: 180px;
+      padding-top: 14px;
```
ami: bahahha. instead of this here, can you remove `height: 43px;` and add `padding: 15px;` to [here](https://github.com/robinjoseph08/pokedextracker.com/blob/be1ad1a6d69483343c2847175a72ecef2268d639/app/styles/main.scss#L94)?

robinjoseph08: w00000w ur gud

PR #177, opened 2016-09-24
feat(reload): add reload component and styles

ami:
⌒°(ᴖ◡ᴖ)°⌒

fixes https://github.com/robinjoseph08/pokedextracker.com/issues/158

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -323,6 +323,18 @@ form {
   }
 }
 
+reload div {
+  text-align: center;
+  padding: 10px 30px;
+  background-color: lighten($brand-primary-light, 15%);
+  font-size: 14px;
+
+  .fa {
+    font-size: 11px;
+    padding-right: 2px;
+  }
+}
+
 @media (max-width: 700px) {
   .not-found h1 {
     font-size: 35px;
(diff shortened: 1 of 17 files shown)


app/styles/main.scss
```diff
@@ -323,6 +323,18 @@ form {
   }
 }
 
+reload div {
+  text-align: center;
+  padding: 10px 30px;
+  background-color: lighten($brand-primary-light, 15%);
```
ami: i can't decide between this color and `$gray-light`. can you try both and let me know what you prefer?

robinjoseph08: im assuming its just `$gray-light` and not `lighten($gray-light, 15%)` cause the latter is pretty much white lol. but i def prefer the yellow for sure 👍

ami: lolz yes that's what i meant. great!

robinjoseph08:
@ami added the functionality! its a tad hard to test locally though (since we dont actually want it to be annoying and stuff when developing). i can either tell you the steps to test it locally ooooooor you can look over the code and then we deploy to staging and we can test it there? whatever works for you

ami:
eh. i'm sure it's fine! MERGGGGGGE

PR #182, opened 2016-10-16
refactor(react): move over reload styles

ami:
ʅฺ(・ω・。)ʃฺ？？

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -40,6 +40,7 @@ a.link {
 
   color: $brand-secondary-light;
   font-weight: 500;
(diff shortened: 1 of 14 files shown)


app/styles/main.scss
```diff
@@ -323,6 +323,22 @@ form {
...
+  padding: 10px 30px;
+  background-color: lighten($brand-primary-light, 15%);
+  font-size: 14px;
+
+  .fa {
+    font-size: 11px;
+    padding-right: 2px;
+  }
+
+  a {
```
ami: can you move this to the `.link` styles?

PR #188, opened 2016-10-22
Small fixes

### LINKED ISSUE #184: Move all errors and loading properties out of state
they don't need to be in the global application state. they could just be in the respective component's local state instead.

### LINKED ISSUE #185: Change all empty elements with self-closing tags
having the full opening/closing tags for empty elements (e.g. `<NavComponent></NavComponent>`) isnt necessary. we can just use the self-closing syntax (e.g. `<NavComponent />`).

robinjoseph08:
( ͒꒪̛ཅ꒪̛ ͒)✧

fixes #184 
fixes #185 

### what

cleanup

### notes

- the main thing that changed was the way errors and loading variables are kept, so just making sure all instances of those (login, register, account, tracker) are still functional would be good.
- i also decided to just remove the state comment that i had in `app/actions/session.js` instead of moving it to the readme like i said i was gonna, mostly because i think [`app/reducers/index.js`](https://github.com/robinjoseph08/pokedextracker.com/blob/master/app/reducers/index.js#L18-L33) is probably the better place to go to for reference.


ami:
lgtm!!

PR #194, opened 2016-10-25
fix(polyfill): add babel-polyfill

### LINKED ISSUE #190: Add Promise polyfill
turns out webpack doesnt provide a promise polyfill like i thought it did. tmyk.

ideally we can replicate it so that we can test that the fix actually works. looks like IE 11 doesnt have `Promise` so a vm should be able to reproduce this.

we could probably use the [`babel-polyfill`](https://babeljs.io/docs/usage/polyfill/) and that would cover it, but raw googling turned up [this gist](https://gist.github.com/Couto/b29676dd1ab8714a818f). just need to evaluate the options.

robinjoseph08:
ฅ(*°ω°*ฅ)

fixes #190 

right now, master is broken for ie 11 (and others). this fixes that! i tested it with my vm! it provides a Promise polyfill and i _think_ a localStorage polyfill too (we've had a few issues with that too).

i also saved the `raw-loader` in the package.json and the shrinkwrap. not sure why it wasnt there before... ¯\\\_(ツ)_/¯

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -549,6 +549,23 @@
       "from": "babel-plugin-transform-strict-mode@>=6.8.0 <7.0.0",
       "resolved": "https://registry.npmjs.org/babel-plugin-transform-strict-mode/-/babel-plugin-transform-strict-mode-6.11.3.tgz"
     },
+    "babel-polyfill": {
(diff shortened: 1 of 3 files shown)


ami:
¯\_(ツ)_/¯ lgtm!

PR #195, opened 2016-10-25
fix(scroll): check if _dex exists

### LINKED ISSUE #191: Check if this.dex exists
there seems to be a possible chain of events that result in an error on [this line](https://github.com/robinjoseph08/pokedextracker.com/blob/master/app/components/dex.jsx#L15). if we just check if `this.dex` exists first, we shouldn't see this error anymore.

robinjoseph08:
(*′☉.̫☉)

fixes #191 

as the issue says, there is a weird occurrence where `this._dex` is `null`. this just checks to make sure that it exists first.

also i dont like the `no-confusing-arrow` rule. its dumb. so i removed it! :radioactive:


ami:
lgtm!

PR #196, opened 2016-10-25
fix(load): wait for DOMContentLoaded if necessary

### LINKED ISSUE #192: Wait for DOMContentLoaded
according to [this article](https://medium.com/@dan_abramov/two-weird-tricks-that-fix-react-7cf9bbdef375#486f) by the creator of redux, he says if you render to a div that's not the `body` element and put your script at the bottom of the `body`, you shouldn't need to wait for this event, but we're doing both of those things and we still see this error. i guess ill try waiting for the event and see if that reduces this error.

robinjoseph08:
(◍˃̶ᗜ˂̶◍)ﾉ”

fixes #192 

as mentioned in the issue, it should wait for the dom to be loaded before trying to load react cause if you try to do so before, it errs out. this should fix that by waiting for `DOMContentLoaded`.

### DIFF
--- a/app/index.jsx
+++ b/app/index.jsx
@@ -6,9 +6,17 @@ import { render }   from 'react-dom';
 import { AppComponent } from './components/app';
 import { Store }        from './stores';
 
-render(
-  <Provider store={Store}>
-    <AppComponent />
-  </Provider>,
-  document.getElementById('root')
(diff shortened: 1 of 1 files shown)


ami:
lgtmmm

PR #200, opened 2016-11-06
feat(profile): connect profile page

ami:
┗|・o・|┛

### DIFF
--- a/app/components/header.jsx
+++ b/app/components/header.jsx
@@ -1,12 +1,9 @@
 import { Component } from 'react';
-import { Link }      from 'react-router';
 import { connect }   from 'react-redux';
 
-import { ProgressComponent } from './progress';
-import { ReactGA }           from '../utils/analytics';
-import { RegionComponent }   from './region';
-import { ShareComponent }    from './share';
-import { setShowShare }      from '../actions/tracker';
+import { ReactGA }        from '../utils/analytics';
+import { ShareComponent } from './share';
+import { setShowShare }   from '../actions/tracker';
 
 export class Header extends Component {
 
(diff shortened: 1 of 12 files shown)


app/components/profile.jsx
```diff
@@ -19,73 +23,60 @@ export class Profile extends Component {
     this.reset();
   }
 
-  componentWillUpdate (props) {
-    this.reset(props);
+  componentWillUpdate ({ params: { username } }) {
```
robinjoseph08: okay so i found a bug where if youre on a profile page (e.g. someone elses like `/u/hiami`) and then you navigate to another profile page directly (e.g. your own by clicking the nav link), it totally flips out (independent of the `NotFoundComponent` bug that i already mentioned). this bug actually already exists in prod for the tracker page, but theres no way to _directly_ navigate to another tracker page from the tracker page (if you edit the url, it causes a refresh which makes it fine).

ill be telling you the lines to edit (and why) for both the profile page and the tracker page.

you should change it from `componentWillUpdate` to `componentWillReceiveProps`. since this function is calling `this.reset()`, it will also be calling `this.setState` (line 35), but turns out ["you cannot call `this.setState()` here. If you need to update state in response to a prop change, use `componentWillReceiveProps()` instead."](https://facebook.github.io/react/docs/react-component.html#componentwillupdate) so there you go.

also, you should change it from `{ params: { username } }` to `nextProps`. the main reason why is because we need to the object intact cause we'll be using it later. ill explain more when we get there.

final result:

```js
componentWillReceiveProps (nextProps) {
```

you should also do this in `tracker.jsx` (line 28)

app/components/profile.jsx
```diff
@@ -19,73 +23,60 @@ export class Profile extends Component {
...
+    if (username !== this.props.params.username) {
```
robinjoseph08: since we havent destructured `nextProps` into `{ params: { username } }`, we no longer have access to `username` directly anymore (your linter will complain about it). so we just need to access it from `nextProps`.

final result:

```js
if (nextProps.params.username !== this.props.params.username) {
```

you should also do this in `tracker.jsx` (line 29)

app/components/profile.jsx
```diff
@@ -19,73 +23,60 @@ export class Profile extends Component {
...
+      this.reset();
```
robinjoseph08: this is gonna be why we needed `nextProps` to be intact. we're actually going to be passing in these new props to the `reset` function. currently, the `reset` function doesnt accept any arguments, but we'll be changing that.

the reason why we need to do that is actually indicated from the previous line. what the `componentWillReceiveProps` (and `componentWillUpdate` for that matter) function is meant for is for you to get a sneak preview as to what the props _will_ be, not what it currently is. `this.props` is still the original props (including the original username and whatnot) and since `reset` uses `this.props`, when we call it, it literally does nothing. it just refetches the already-fetched user.

so. that means we need to give this sneak preview of the new props to the reset function and conditionally use that instead of `this.props`. if that doesnt make sense, let me know!!

final result:

```js
this.reset(nextProps);
```

you should also do this in `tracker.jsx` (line 30)

app/components/profile.jsx
```diff
@@ -19,73 +23,60 @@ export class Profile extends Component {
...
+    }
   }
 
-  reset (props) {
-    const { checkVersion } = props || this.props;
+  reset () {
```
robinjoseph08: since we're sometimes passing in new props into this function we need to be able to utilize it, so we'll be listing `props` as an argument. if `this.reset` is called without any arguments (e.g. `this.reset()` vs `this.reset(nextProps)`, `props` will be `undefined`, and we'll be utilizing this fact in a bit.

final result:

```js
reset (props) {
```

you should also do this in `tracker.jsx` (line 34)

app/components/profile.jsx
```diff
@@ -19,73 +23,60 @@ export class Profile extends Component {
...
+    const { checkVersion, params: { username }, retrieveUser, setCurrentUser, setShowShare } = this.props;
```
robinjoseph08: in this line, we're pulling out all of these variables and functions out of `this.props`, including `username`. like i said before, when we call this in `componentWillReceiveProps` this wont be using all the new stuff, itll still be using the old props. so we ned to conditionally use the new props that are being passed in. thats super easy to do. we just need to replace `this.props` with `props || this.props`. so it'll try to use `props` first but if it's `undefined`, itll default to using `this.props`. easy peasy.

final result:

```js
const { checkVersion, params: { username }, retrieveUser, setCurrentUser, setShowShare } = props || this.props;
```

you should also do this in `tracker.jsx` (line 35)

app/components/profile.jsx
```diff
@@ -94,13 +85,16 @@ export class Profile extends Component {
 
 }
 
-function mapStateToProps ({ session }) {
-  return { session };
+function mapStateToProps ({ users }, { params: { username } }) {
```
robinjoseph08: see the comma we use here? thats splitting up the two arguments that're being passed in (see docs [here](https://github.com/reactjs/react-redux/blob/master/docs/api.md#arguments), though you dont have to). the first argument is the global state, the second is the react component's original props (the ones that're being passed to it). this part is a bit tricky, but we actually want to get the current user using `currentUser`, not `username`. the reason for this is because we want `user` to change as soon as possible, so that begs the question, which one updates first, `currentUser` or `username`. well `currentUser` changes when we call `setCurrentUser` which is in the `reset` function and `username` will update when the routing navigation fully completes, which is after `componentWillReceiveProps` (and `reset`) is finished running. so that means `currentUser` changes first. thats why we should use that. and since `currentUser` is in the global state, we don't even need the second arg anymore.

final result:

```js
function mapStateToProps ({ currentUser, users }) {
```

you should also do this in `tracker.jsx` (line 84)

app/components/profile.jsx
```diff
@@ -94,13 +85,16 @@ export class Profile extends Component {
...
+  return { user: users[username] };
```
robinjoseph08: since we're no longer using `username`, we just need to change this to `currentUser` instead.

final result:

```js
return { user: users[currentUser] };
```

you should also do this in `tracker.jsx` (line 85)

and thats it! hopefully that all made sense. if it didn't feel free to ask questions. id like for you to understand what the problem was too! now if you try to navigate to another profile page, it should work juuuust fine 👍

ami:
@robinjoseph08 updated! also made some small style changes to accommodate for long dex titles (noticed that they can be up to 300 characters)

app/components/header.jsx
```diff
@@ -33,37 +30,29 @@ export class Header extends Component {
...
-        </h2>
-
-        <RegionComponent />
-
-        <div className="percentage">
-          <ProgressComponent />
-          <RegionComponent mobile={true} />
-        </div>
-      </header>
+      <div>
```
robinjoseph08: unless it's needed for stylez, this `div` isnt necessary since the `h1` can act as the container

PR #203, opened 2016-11-06
feat(dexes): add dex settings to signup

ami:
(੭ु´͈ ᐜ `͈)੭ु⁾⁾

### what
- add dex settings to signup
- use shiny sprites on tracker page if dex is shiny

### notes
- i added the dex stuff beneath the user stuff in one column, which i think is fine - but it's a bit weird when you submit the form and there's an error because you have to scroll back up to see the error. any way for us to auto-scroll up to the error? alternatively, i can refactor the register page to be two columns instead if you think that's better
- after registration, you get redirected to the profile page. do you think that's preferred or should we send you directly to your first dex?

### DIFF
--- a/app/components/login.jsx
+++ b/app/components/login.jsx
@@ -19,10 +19,10 @@ export class Login extends Component {
   }
 
   componentWillMount () {
-    const { checkVersion, redirectToTracker, session } = this.props;
+    const { checkVersion, redirectToProfile, session } = this.props;
 
     if (session) {
-      redirectToTracker(session.username);
+      redirectToProfile(session.username);
     }
 
     checkVersion();
@@ -53,19 +53,21 @@ export class Login extends Component {
           <div className="form">
             <h1>Login</h1>
             <form onSubmit={this.login}>
-              <AlertComponent message={error} type="error" />
-              <div className="form-group">
-                <label htmlFor="username">Username</label>
-                <input className="form-control" ref={(c) => this._username = c} name="username" id="username" type="text" required placeholder="ashketchum10" maxLength="20" autoComplete="off" autoCorrect="off" autoCapitalize="off" spellCheck="false" />
-                <i className="fa fa-asterisk" />
+              <div className="form-column">
+                <AlertComponent message={error} type="error" />
+                <div className="form-group">
+                  <label htmlFor="username">Username</label>
+                  <input className="form-control" ref={(c) => this._username = c} name="username" id="username" type="text" required placeholder="cabrioles" maxLength="20" autoComplete="off" autoCorrect="off" autoCapitalize="off" spellCheck="false" />
+                  <i className="fa fa-asterisk" />
+                </div>
+                <div className="form-group">
+                  <label htmlFor="password">Password</label>
+                  <input className="form-control" ref={(c) => this._password = c} name="password" id="password" type="password" required placeholder="••••••••••••" maxLength="72" />
+                  <i className="fa fa-asterisk" />
+                </div>
+                <button className="btn btn-blue" type="submit">Let's go! <i className="fa fa-long-arrow-right" /></button>
+                <p>Don't have an account yet? <Link className="link" to="/register">Register here</Link>!</p>
               </div>
-              <div className="form-group">
-                <label htmlFor="password">Password</label>
-                <input className="form-control" ref={(c) => this._password = c} name="password" id="password" type="password" required placeholder="••••••••••••" maxLength="72" />
-                <i className="fa fa-asterisk" />
-              </div>
-              <button className="btn btn-blue" type="submit">Let's go! <i className="fa fa-long-arrow-right" /></button>
-              <p>Don't have an account yet? <Link className="link" to="/register">Register here</Link>!</p>
             </form>
           </div>
         </div>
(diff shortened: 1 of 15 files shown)


app/components/register.jsx
```diff
@@ -53,29 +56,67 @@ export class Register extends Component {
...
               <div className="form-group">
                 <label htmlFor="friend_code">Friend Code</label>
-                <input ref={(c) => this._friend_code = c} name="friend_code" id="friend_code" type="text" placeholder="XXXX-XXXX-XXXX" onChange={(e) => this._friend_code.value = friendCode(e.target.value)} />
+                <input className="form-control" ref={(c) => this._friend_code = c} name="friend_code" id="friend_code" type="text" placeholder="XXXX-XXXX-XXXX" onChange={(e) => this._friend_code.value = friendCode(e.target.value)} />
+              </div>
+
+              <h2>
+                First Dex Info
+                <div className="tooltip">
+                  <i className="fa fa-plus-circle" />
```
robinjoseph08: is there a reason you picked a `+` over a `?`? it kinda seemed like it was a button to make another dex at first

app/utils/pokemon.js
```diff
@@ -4,8 +4,13 @@ export function htmlName (name) {
   return { __html: name.replace('♀', '<i class="fa fa-venus"></i>').replace('♂', '<i class="fa fa-mars"></i>') };
 }
 
-export function iconClass (id) {
-  return `pkicon pkicon-${padding(id, 3)}`;
+export function iconClass (id, shiny) {
+  const idPadded = padding(id, 3);
```
robinjoseph08: can you call this `paddedId` instead?

app/utils/pokemon.js
```diff
@@ -4,8 +4,13 @@ export function htmlName (name) {
...
+
+  if (shiny) {
+    return `pkicon pkicon-${idPadded} color-shiny`;
+  }
+  return `pkicon pkicon-${idPadded} test`;
```
ami: oopz! nope. that was me testing hehe.

app/components/register.jsx
ami: i think i just ported it over from the tooltips we have already on the tracker page:
image

but in that context, the `+` makes sense while it doesn't rly here. will change to a `?`!

ami:
- register page:
image
this is what it might look like. do you think it's better, considering that it's uneven? but ya, we will still have to solve the problem for mobile (b/c on mobile, we would definitely turn it into 1 column.

- will add redirect to dex 👍 
- i think checkboxes are good for the future "stackable" features, but would like to keep the radio for now when we only have shiny (if only to take up more space visually)
- will look into adding shiny stars in a different PR!
- i did do that intentionally b/c i like having the larger click-area for the link, but i can remove it if you think it's weird! was just thinking that some dex titles might be very short
- achhh i changed some of those styles because it was uncentered on my laptop, but now i'm thinking maybe it's a retina/non-retina thing? like maybe when i first made them it was on my work laptop but now on my personal (non-retina) it looks off-center? if so, we should probably optimize for retina since that's what most people have now.

robinjoseph08:
- yeah even being uneven, i think its better. what do you think?
- 👌 to the 📻 
- i think its a little weird mostly cause its an unbounded click area. like, if there was a border around whole thing and the whole thing was clickable, then itd be fine. especially with the edit button over on the right, it was unexpected to see the space in between a link to the thing on the left. does that make sense?
- thats weird that retina v non-retina are different. but idk if retina is more prevalent with our audience. according to GA, windows is the top OS by a significant margin (the number of sessions for android is 1/4 the number for windows). with that said, ill defer to your judgement as to what to do.

  image

ami:
- 👍  plus, there will one day be more options that will even it out
- 👍 
- 👍 
- interesting. ok ideally i'll try to figure out why it's weird for both and try to make it best for both, but if not, will optimize for the non-retina case.

robinjoseph08:
also we should rebase this branch (and the feature branch) with master! i can do it if you want, but we'd need to do it carefully so no changes get force-pushed away. ooooor you can just do it since youre a GIT PRO

ami:
@robinjoseph08 updated! i couldn't actually figure out how to add the redirect directly to the first dex.. so if you could add that here that would be great! i know you have to retrieve the user to get the slug but i no understands javascript 😕 

also, can you add jumping up to the error on the mobile register page?

app/components/register.jsx
```diff
@@ -53,31 +56,79 @@ export class Register extends Component {
...
+                <div className="form-column">
+                  <h2>
+                    First Dex Info
+                    <div className="tooltip">
+                      <i className="fa fa-question-circle" />
+                      <span className="tooltip-text">You can track multiple dexes on our app! This sets the settings for the first dex on your account.</span>
+                    </div>
+                  </h2>
+                  <div className="form-group">
+                    <label htmlFor="dex_name">Name</label>
```
robinjoseph08: is there a reason you called it `Name` instead of `Title`?

app/components/login.jsx
```diff
@@ -53,19 +53,21 @@ export class Login extends Component {
           <div className="form">
             <h1>Login</h1>
             <form onSubmit={this.login}>
```
robinjoseph08: should this have `className="form"`?

robinjoseph08: wait nope, im dumb. ignooooore

app/components/register.jsx
robinjoseph08: okie im gonna change it to title to stay consistent!

robinjoseph08:
@ami done and done!

for the redirecting, you're right, we didn't have enough info to be able to redirect immediately (since we need the slug) but making a request to retrieve the user seems a bit wasteful (since that info would immediately be discarded). so instead we're just using the same module to create the slug so that we can avoid that http request!

ami:
sweet. LGTM!!

PR #204, opened 2016-11-07
refactor(dexes): use the dexes API for the tracker page

robinjoseph08:
(ꐦ°᷄д°᷅)

this is the same commit from #199, but merging to master instead of the feature branch, except it's also passing `dex` when saving/deleting captures. as i was testing #203, i created a shiny dex for `/u/hiami` which was the third dex for that account but every time i marked a pokemon as captured, it didnt actually mark it as captured in the db. this is cause the endpoint just picks a random dexs to create the captures for lol. but for the api to use a dex id thats being passed in, this needs to be in master.

you should just make sure that everything still works as expected (particularly users with only one dex)!

### DIFF
--- a/app/actions/capture.js
+++ b/app/actions/capture.js
@@ -5,37 +5,37 @@ import { checkVersion } from './utils';
 export const MARK_CAPTURED = 'MARK_CAPTURED';
 export const SET_CAPTURES  = 'SET_CAPTURES';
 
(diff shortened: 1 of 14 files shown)


robinjoseph08:
yup! its only supposed to work for single dexes for now. once this gets deployed, i can modify the api to work with multiple dexes

PR #205, opened 2016-11-14
feat(nav): update nav for multiple dexes

ami:
(灬㋭ ਊ ㋲灬)

updates nav and home links for multiple dexes! i THINK it mostly works but there may be random edge cases i'm not sure about ;____;

### DIFF
--- a/app/components/nav.jsx
+++ b/app/components/nav.jsx
@@ -2,8 +2,9 @@ import { Component } from 'react';
 import { Link }      from 'react-router';
 import { connect }   from 'react-redux';
 
-import { ReactGA }  from '../utils/analytics';
-import { setToken } from '../actions/session';
+import { ReactGA }      from '../utils/analytics';
+import { retrieveUser } from '../actions/user';
+import { setToken }     from '../actions/session';
 
 export class Nav extends Component {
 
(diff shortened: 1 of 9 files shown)


app/components/nav.jsx
```diff
@@ -15,15 +16,53 @@ export class Nav extends Component {
...
+  }
+
+  componentWillMount () {
+    this.reset();
+  }
+
+  reset (props) {
+    const { retrieveUser, session } = props || this.props;
+
+    if (session) {
```
ami: wrapped this here to only do this stuff if `session`... not sure if this is the best place/way to do that.

app/components/nav.jsx
```diff
@@ -15,15 +16,53 @@ export class Nav extends Component {
...
+
   render () {
-    const { session } = this.props;
+    const { session, user } = this.props;
+    const { loading } = this.state;
+
+    if (loading) {
+      return (
+        <nav>
+          <Link to="/">Pokédex Tracker</Link>
```
ami: originally i had this whole thing as `null`, but changed to this to prevent as much "flickering" as possible. maybe there's a better way to avoid that with the whole better loading thing... idk

robinjoseph08:
found a few weird edge cases, mostly caused by both the nav and the tracker trying to load the user into the same place in the global state. pushed a commit that saves the nav user to `sessionUser` instead so hopefully it fixes everything. we should test some more though. put your qa skillz to the test!

robinjoseph08:
poops, i meant to add that the alignment of the icons in the nav is a bit weird if there are many with diff lengths:

image

do you think so too or am i crazy? for some reason i think the icons should be aligned

ami:
tested a bunch of cases, lgtm! when i was designing, i didn't have a strong preference between aligned or not so i went with not cause i thought it looked slightly more interesting. but since i don't really feel strong, i updated it to be aligned!

screenshot 2016-11-17 23 23 42

robinjoseph08:
meh, i think it looks slightly more off with the rest of them:

image

PR #206, opened 2016-11-16
feat(dex): add create dex modal

robinjoseph08:
♫ꉂ (๑¯ਊ¯)σ

- need you do work your stylez magic on this bad boy. the two main class you need to work with are `.modal` and `.modal-overlay`.
- we should probs pull out this form since we use it here and register... oh well.
- i didnt add a loading var to this component since i was going off login/register which dont also, but we can add one if you think we should. i kinda think we should add them to login/register too

### DIFF
--- a/app/components/create-dex.jsx
+++ b/app/components/create-dex.jsx
@@ -1,0 +1,106 @@
+import { Component } from 'react';
+import Modal         from 'react-modal';
+import { connect }   from 'react-redux';
+import { push }      from 'react-router-redux';
+
+import { AlertComponent } from './alert';
+import { ReactGA }        from '../utils/analytics';
+import { createDex }      from '../actions/dex';
+
+export class CreateDex extends Component {
+
+  constructor (props) {
+    super(props);
+    this.state = { error: null };
+  }
+
+  scrollToTop () {
+    if (this._form) {
+      this._form.scrollTop = 0;
+    }
+  }
+
+  onSubmit = (e) => {
+    e.preventDefault();
+
+    const { createDex, redirect, session } = this.props;
+    const title = this._title.value;
+    const shiny = this._shiny.checked;
+    const generation = this._generation.value;
+    const payload = {
+      username: session.username,
+      payload: { title, shiny, generation }
+    };
+
+    this.setState({ ...this.state, error: null });
+
+    createDex(payload)
+    .then((dex) => {
+      ReactGA.event({ action: 'create', category: 'Dex' });
+
+      redirect(`/u/${session.username}/${dex.slug}`);
+    })
+    .catch((err) => {
+      this.setState({ ...this.state, error: err.message });
+      this.scrollToTop();
+    });
+  }
+
+  render () {
+    const { isOpen, onRequestClose } = this.props;
+    const { error } = this.state;
+
+    return (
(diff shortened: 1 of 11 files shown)


app/components/create-dex.jsx
```diff
@@ -0,0 +1,108 @@
...
+                <label>
+                  <input ref={(c) => this._shiny = c} type="radio" name="type" />
+                  <span className="radio-custom"><span></span></span>Shiny
+                </label>
+              </div>
+            </div>
+            <button className="btn btn-blue" type="submit">Create <i className="fa fa-long-arrow-right" /></button>
+          </form>
+        </div>
+        <p><Link className="link">Go Back</Link></p>
```
ami: added this link to "close" the modal - can you add the functionality?

app/components/create-dex.jsx
```diff
@@ -0,0 +1,108 @@
...
+import { Link }      from 'react-router';
+import Modal         from 'react-modal';
+import { connect }   from 'react-redux';
+import { push }      from 'react-router-redux';
+
+import { AlertComponent } from './alert';
+import { ReactGA }        from '../utils/analytics';
+import { createDex }      from '../actions/dex';
+
+export class CreateDex extends Component {
```
ami: thoughts on calling this file `dex-create` and component `DexCreate`? it's more consistent with `dex` and `dex-preview` and `Dex` and `DexPreview` - plus then all the dex-related things are listed next to each other.

robinjoseph08:
1. yeah we can deal with it separately
2. mines not showing up alphabetically? its sorted by date created desc for me:

image

i can make it asc though if you think thats better.

3. i made it say that since technically its the url that needs to be unique. like i can have `Living Dex` and try to create `!!LIVING DEX!!` and it would error, but that title doesnt exist. im open to suggestions though

robinjoseph08:
1. without a 1, github starts the numbering at 1, even if i start it with 2. soooo
2. lol yeah i can change it to asc (api change). i think at the bottom is better like what we have
3. yeah i think both would be helpful. ill change the error message (api change) and you add the tooltip?
4. oy. i think that could be a v2 thing. its really only necessary once pokebank is available, right?

ami:
@robinjoseph08 review pls, added url preview (as opposed to a tooltip)

app/components/register.jsx
```diff
@@ -106,7 +106,7 @@ export class Register extends Component {
                   </h2>
                   <div className="form-group">
                     <label htmlFor="dex_title">Title</label>
```
robinjoseph08: thoughts on adding the url preview here too? not really to inform the user about conflicting urls, but just cause its cool looking

robinjoseph08: The username hasn't been finalized but we could use what they've already typed into the username input, right? And just default it to ashketchum10 if its empty, like what we do with Living Dex for title?

ami: ya, but in that case i think it's weird because with Living Dex - that'd be something completely reasonable to name your dex, while ashketchum10 isn't (except for that one user lols). because of that, i think it'd be more confusing than anything to put it there. and since - like you mentioned - it doesn't provide much benefit on the register page other than looking cool, i'd rather not introduce a possible point of confusion without much utility.

robinjoseph08: Hmmm gotcha. Yeah I can see that could be confusing. Yeah we can just table it for now :+1:

PR #207, opened 2016-11-20
refactor(info): generate urls on the front-end and add spaces

robinjoseph08:
(⋆ॢʾ ˙̫̮ ʿ⋆ॢ)

stop relying on the returned props so that we can generate them based on dex. this is necessary for gen vii support. this is blocking https://github.com/robinjoseph08/api.pokedextracker.com/pull/69 so we should deploy this as soon as its merged!

i also noticed that we were missing some spaces. some examples of why we need it:

image

image

image

i dont think adding these spaces will hurt if they arent needed, right?

### DIFF
--- a/app/components/evolutions.jsx
+++ b/app/components/evolutions.jsx
@@ -7,13 +7,13 @@ export function EvolutionsComponent ({ evolutions }) {
 
     switch (evolution.trigger) {
       case 'level':
-        trigger = <span>Level Up {evolution.level ? `to ${evolution.level}` : ''}</span>;
+        trigger = <span>Level Up {evolution.level ? `to ${evolution.level} ` : ''}</span>;
         break;
       case 'stone':
-        trigger = <span>{capitalize(evolution.stone)} Stone</span>;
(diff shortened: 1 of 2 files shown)


robinjoseph08:
No, this doesn't have anything to do with any of the new stuff we're adding. The spaces are necessary even for master. Notes is only in a tooltip if and only if the number of evos is >= 3 (so Eevee and Tyrogue). Otherwise, it's just displayed like normal. If you look at the screenshots, you can see how there are certain words that are together without spaces between them (`Stoneduring`, `Breedholding`, `Stone(male)`, etc). This is just because of how the elements are conditionally constructed. Adding the spaces in these places should catch all possible outcomes and make everything look right. Does that make sense?

ami:
ooOooOooooooooooooooic. weird, so i guess this just came through when we moved to react but we didn't notice, because that's what moved those things into the different conditional structure?

robinjoseph08:
Right, I'm pretty sure that's when it was introduced. MEEEERGING

PR #210, opened 2016-11-27
feat(dexes): add shiny dex indicator

ami:
ヾ(０∀０*★)ﾟ*･.｡

- adds a shiny indicator for shiny dexes! inspired by pokemon's [own way of indicating shiny pokemon](http://i.imgur.com/ndMSJ3e.jpg)

### DIFF
--- a/app/components/dex-indicator.jsx
+++ b/app/components/dex-indicator.jsx
@@ -1,0 +1,11 @@
+export function DexIndicatorComponent ({ dex }) {
+  if (!dex || !dex.shiny) {
+    return null;
+  }
+
+  return (
+    <div className="dex-indicator" title="shiny">
+      <i className="fa fa-star" />
+    </div>
+  );
+}
--- a/app/components/dex-preview.jsx
+++ b/app/components/dex-preview.jsx
@@ -1,12 +1,14 @@
 import { Link }    from 'react-router';
 import { connect } from 'react-redux';
 
-import { ProgressComponent } from './progress';
+import { DexIndicatorComponent } from './dex-indicator';
+import { ProgressComponent }     from './progress';
 
 export function DexPreview ({ dex, username }) {
   return (
     <div className="dex-preview">
       <div className="dex-preview-header">
+        <DexIndicatorComponent dex={dex} />
         <h3><Link to={`/u/${username}/${dex.slug}`} className="link">{dex.title}</Link></h3>
         <Link to=""><i className="fa fa-pencil" /></Link>
       </div>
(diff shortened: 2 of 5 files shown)


app/components/dex-preview.jsx
```diff
@@ -4,9 +4,20 @@ import { connect } from 'react-redux';
 import { ProgressComponent } from './progress';
 
 export function DexPreview ({ dex, username }) {
+  let dexIndicator = null;
+
+  if (dex.shiny) {
+    dexIndicator = (
+      <div className="dex-indicator">
+        <i className="fa fa-star" />
```
robinjoseph08: thoughts on having a `title` attr just in case someone is new to it all and wants to know what the red star means?

ami:
@robinjoseph08 updated!

app/components/dex-indicator.jsx
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux';
+
+export function DexIndicator ({ dex, profile }) {
+
+  if ((dex && !dex.shiny) || profile) {
```
robinjoseph08: hmm is this `|| profile` necessary? cause wont `dex` be `undefined` if `profile` is true?

app/components/dex-indicator.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+  return (
+    <div className="dex-indicator" title="shiny">
+      <i className="fa fa-star" />
+    </div>
+  );
+}
+
+function mapStateToProps ({ currentUser }) {
+  return { username: currentUser };
```
robinjoseph08: you don't use `username` in the component so its unnecessary. which makes `mapStateToProps` unnecessary. which makes `connect` unnecessary.

app/components/dex-indicator.jsx
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux';
+
+export function DexIndicator ({ dex, profile }) {
+
```
robinjoseph08: extra new line

app/components/dex-indicator.jsx
ami: right, that's why `|| profile` is necessary - i want this expression to evaluate to true if we're on the `profile`. the logic is: render nothing if i'm a dex and the dex is not shiny, or if i'm on the profile page (because this component is used in the `HeaderComponent` which is used on both the profile page and the dex page, but in that case i only want it to show on the dex page). is there a better way to write that out?

robinjoseph08: I think `if (!dex || !dex.shiny) {` will do the same thing. Could you try that?

ami:
@robinjoseph08 yayyy it worked! wanna take another look?

PR #224, opened 2016-12-12
fix(nav): fix dropdown for mobile

ami:
¯\_╏ ՞ ︿ ՞ ╏_/¯

fixes https://github.com/pokedextracker/pokedextracker.com/issues/223

quick fix to allow the dropdown to work on mobile. you can't "un"click it to make it go away, but at least it works now! not sure if there's an "actual" fix other than just making a mobile-only version of the dropdown that is based off clicks instead of hover.


ami:
I believe this is how bootstrap handles it (lob dashboard is the same as this). Everywhere I read just recommends changing to a click (hamburger menu) for mobile.

PR #238, opened 2017-01-28
Shrinkwrap cleanup

SimenB:
Using newer npm, it adds into the shrinkwrap whether a dep is a dev and/or an optional dep.

Also added missing babel-core to package.json.

I use yarn, which is why it wasn't installed. Have you considered using yarn? Shrinkwrap by default and much speedier 😄


SimenB:
I _think_ this change means the hack for fsevents is unneeded btw, it's marked as optional

robinjoseph08:
@SimenB what do you mean by "missing babel-core"? Is there a dependency that we currently have that has it as a peer dependency? I'm wondering why this hasn't caused any problems before?

SimenB:
@robinjoseph08  Yes, that's exactly it. It hasn't caused problems because npm treats the shrinkwrap file as the source of truth, so you got babel-core installed from there. My guess would be that you generated the shrinkwrap file using npm@2, which automatically installed peer deps.

I can rebase this PR if you want? Might be better if you just went for a yarn lockfile, straight away, though 😄 

If you wanna wait on the lockfile, you should at least update the shrinkwrap using a newer npm, so you get the `optional` and `dev` properties this PR proved (again, I can rebase if you want)

SimenB:
Isee from you OP in #253 that you'll go for Yarn, so I'll just close this 😄

PR #239, opened 2017-01-28
Webpack 2

SimenB:
Webpack 2 is released and stable, so why not use it 😄 

This builds atop of #238 

Had to manually dedupe a bunch of stuff since shrinkwraps are hard. Wanna reiterate the point that yarn is easier to use 😁


robinjoseph08:
Thanks for wanting to keep our dependencies up to date! Last time I checked, v2 was still in beta so it's great that it was officially released. Looks like there were a few things missing during the upgrade process that I found when verifying you PR, so I went ahead and took care of it in [this commit](https://github.com/pokedextracker/pokedextracker.com/commit/486b86427cf2953aa1dd7c5e421af3d644df657c). I also upgraded all of the other packages in this repo so that we can make the switch over to yarn without any side effects! Thanks for your support with it!

PR #240, opened 2017-01-28
Hide caught pokemon

### LINKED ISSUE #136: Have a view of only uncaught pokemon
as you start getting closer to 100%, it gets difficult to keep scrolling up and down to find the pokemon you have left. if there was a way to just view all uncaught pokemon, it'll make the home stretch much easier

SimenB:
Fixes #136 

This is a quick implementation that just hides the ones that are caught.

image

image

### DIFF
--- a/app/actions/filters.js
+++ b/app/actions/filters.js
@@ -1,0 +1,5 @@
+export const TOGGLE_SHOW_CAUGHT_POKEMON = 'TOGGLE_SHOW_CAUGHT_POKEMON';
+
+export function toggleShowCaughtPokemon () {
+  return { type: TOGGLE_SHOW_CAUGHT_POKEMON };
+}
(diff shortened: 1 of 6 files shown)


SimenB:
I want to be able to filter out pokemon only available in a given game (e.g. just OR pokemons). That's why I went with a whole section for filter.
Not sure if your API supports it or if I'd have to tweak it for that feature though 😄

ami:
thanks for the contribution, @SimenB! while this is definitely functionality we want, we're hoping to implement it more fully, with [additional filtering and perhaps even some kind of search](https://github.com/pokedextracker/pokedextracker.com/issues/136). additionally, the UI/UX isn't up to par with what we expect for this product.

usually if there's change you want to make, its probably a good idea to make an issue for it to see how it aligns with our plans for the site. if an issue already exist (like in this case), it's helpful to comment on it so that we can make sure it isn't something we're already in the middle of working on. if it isn't, and there's room for an external contribution, then a comment will helps us guide you on what we were thinking the functionality and ui/ux. this will help us both out to make sure were all on the same page before any time is spent!

PR #241, opened 2017-01-29
feat(create): add region to create dex

ami:
☆*\~ﾟ⌒(‘-‘*)⌒ﾟ\~*☆

### DIFF
--- a/app/components/dex-create.jsx
+++ b/app/components/dex-create.jsx
@@ -12,7 +12,17 @@ export class DexCreate extends Component {
 
   constructor (props) {
     super(props);
-    this.state = { error: null };
+    this.state = { error: null, generation: 7, region: 'national' };
+  }
+
+  onChange = (e) => {
+    const generation = parseInt(e.target.value);
+
+    if (generation === 6) {
+      this.setState({ region: 'national' });
+    }
+
+    this.setState({ generation });
   }
 
   scrollToTop () {
@@ -22,7 +32,7 @@ export class DexCreate extends Component {
   }
 
   onRequestClose = () => {
-    this.setState({ error: null, url: null });
+    this.setState({ error: null, url: null, generation: 7, region: 'national' });
     this.props.onRequestClose();
   }
 
@@ -30,10 +40,9 @@ export class DexCreate extends Component {
     e.preventDefault();
 
     const { createDex, redirect, session } = this.props;
+    const { generation, region } = this.state;
     const title = this._title.value;
     const shiny = this._shiny.checked;
-    const generation = this._generation.value;
-    const region = parseInt(generation) === 7 ? 'alola' : 'national';
     const payload = {
       username: session.username,
       payload: { title, shiny, generation, region }
@@ -55,7 +64,7 @@ export class DexCreate extends Component {
 
   render () {
     const { isOpen, session } = this.props;
-    const { error, url } = this.state;
+    const { error, generation, region, url } = this.state;
 
     return (
       <Modal className="modal" overlayClassName="modal-overlay" isOpen={isOpen} onRequestClose={this.onRequestClose}>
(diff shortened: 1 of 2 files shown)


app/components/dex-create.jsx
```diff
@@ -12,7 +12,16 @@ export class DexCreate extends Component {
...
   constructor (props) {
     super(props);
-    this.state = { error: null };
+    this.state = { error: null, generation: '7', region: 'national' };
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange (event) {
+    if (this.state.generation === '7' && this.state.region === 'alola') {
```
robinjoseph08: instead looking at the current state, you should look at what we're changing the state to. itll make the logic easier to understand in the future

```js
onChange = (e) => {
  const generation = parseInt(e.target.value);

  if (generation === 6) {
    this.setState({ region: 'national' });
  }

  this.setState({ generation });
}
```

^ this _should_ be the same logic, but you should double check

app/components/dex-create.jsx
```diff
@@ -32,8 +41,8 @@ export class DexCreate extends Component {
     const { createDex, redirect, session } = this.props;
     const title = this._title.value;
     const shiny = this._shiny.checked;
-    const generation = this._generation.value;
-    const region = parseInt(generation) === 7 ? 'alola' : 'national';
+    const generation = this.state.generation;
+    const region = this.state.region;
```
robinjoseph08: instead of using `this.state` directly, we should destructure it just like what we do for `this.props` above. thats what we do [elsewhere](https://github.com/pokedextracker/pokedextracker.com/blob/master/app/components/account.jsx#L73-L74)

```js
const { createDex, redirect, session } = this.props;
const { generation, region } = this.state;
```

app/components/dex-create.jsx
```diff
@@ -71,13 +80,28 @@ export class DexCreate extends Component {
             </div>
             <div className="form-group">
               <label htmlFor="generation">Generation</label>
-              <select className="form-control" ref={(c) => this._generation = c} defaultValue="7">
+              <select className="form-control" onChange={this.handleChange} value={this.state.generation}>
```
robinjoseph08: since we destructure `generation` on line 67, we can just use `generation` instead of `this.state.generation`

app/components/dex-create.jsx
```diff
@@ -71,13 +80,28 @@ export class DexCreate extends Component {
...
                 <option value="7">Seven</option>
                 <option value="6">Six</option>
               </select>
               <i className="fa fa-chevron-down" />
             </div>
             <div className="form-group">
+              <label htmlFor="region">Regionality</label>
+              <div className="radio">
+                <label>
+                  <input type="radio" name="region" checked={region === 'national'} value="national" onChange={() => this.setState({ ...this.state, region: 'national' })} />
```
robinjoseph08: i know we have it in several other places (cause i thought it was necessary), but you're `handleChange` function showed me that `...this.state` isn't necessary.

i thought `this.setState` overwrote the value of `this.state` so whenever we called it, we had to make sure we included all the other props (like `loading` and `error` and stuff) and thats what `...this.state` is doing. but `this.setState` actually takes what you pass into it and merges it with the current state so we dont have to do it.

app/components/dex-create.jsx
```diff
@@ -71,13 +80,28 @@ export class DexCreate extends Component {
...
+                  <span className="radio-custom"><span /></span>National
+                </label>
+              </div>
+              <div className={`radio ${generation === '6' ? 'disabled' : ''}`}>
+                <label>
+                  <input type="radio" name="region" checked={region === 'alola'} disabled={generation === '6'} value="alola" onChange={() => this.setState({ ...this.state, region: 'alola' })} />
```
robinjoseph08: should we maybe have a `title` attr if it's disabled to give a tiny explanation why its disabled?

ami: i thought about it, but wasn't sure what to put though? Could be something like `Not yet supported` but that doesn't give you any more information than what the disabled state already does, ya know? but not opposed to it.

robinjoseph08: yeah idk. maybe something like `Regional only support for gen 7` or something?

ami:
updated!

PR #245, opened 2017-01-30
feat(boxes): take pokemon.box into account when grouping

robinjoseph08:
⤴︎ ε=ε=(ง ˃̶͈̀ᗨ˂̶͈́)۶ ⤴︎

- group the boxes according to `pokemon.box`
- use the correct title for the box
- use `pokemon.form` for the sprite

this should be the last pr i think! just gotta make it look pertty


robinjoseph08:
@ami updated! not super happy with it cause now we can't break up non-national dexes according to box, but we dont need that now, so we'll deal with it later

robinjoseph08:
lgtm!

PR #246, opened 2017-02-04
Replace npm shrinkwrap with yarn lockfile

SimenB:
I mentioned it in a few of my other PRs, and I thought I'd just provide the PR so you can merge it.

With Yarn being built around the lockfile, it's very easy to manage, while npm's shrinkwrap was an afterthought and has lots of issues (with deduping, upgrading and, as you noticed, optional dependencies).
You never prune, dedupe or make sure the shrinkwrap file is written, it's all handled for you. And if someone changes the package.json without updating the lockfile, it's synced the next time, as it's the source of truth, unlike shrinkwrap which is an absolute copy of how it's laid out on disk at the time it's written.

Travis automatically uses yarn if it sees a lockfile, so no changes necessary there.

One thing to note is that the old shrinkwrap wasn't respected when generating this lockfile. But from my tests, everything still works like they should 😄 

This closes #238 if merged.


SimenB:
This closes #238 if merged 😄

robinjoseph08:
I've actually had plans to use yarn for this repo since it first came out. It seems way better than npm in so many different aspects that have been pain points for me. I actually should've created an issue for it, but it slipped my mind.

The main reason I haven't converted over just yet is because, like you said, yarn can't use a pre-existing `npm-shrinkwrap.json` to generate the `yarn.lock` file. This inadvertently upgrades several dependencies (both direct and indirect ones), and the whole reason we have the shrinkwrap file is to avoid that.

The way that I hope to actually do the conversion is to go through and update all of our dependencies so that when we switch, there's no chance of accidental updates. It also allows us to monitor all of the potential changes and it makes for a better QA process when we do the individual updates.

I'm gonna close this PR for now, but I'll be working on updating the deps so that we can ultimately convert to yarn.

PR #250, opened 2017-02-05
Add name of pokemon on hover in evolution chain

SimenB:
Looks like 
image

### DIFF
--- a/app/components/evolution-family.jsx
+++ b/app/components/evolution-family.jsx
@@ -11,7 +11,7 @@ export function EvolutionFamily ({ dex, family, setCurrentPokemon }) {
   if (family.pokemon.length > 1) {
     column1 = (
       <div className="evolution-pokemon-column">
-        {family.pokemon[1].map((pokemon) => <a key={pokemon.id} onClick={() => setCurrentPokemon(pokemon.id)}>
+        {family.pokemon[1].map((pokemon) => <a key={pokemon.id} onClick={() => setCurrentPokemon(pokemon.id)} title={pokemon.name}>
           <i className={iconClass(pokemon, dex)} />
         </a>)}
       </div>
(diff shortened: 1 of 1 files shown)


app/components/evolution-family.jsx
```diff
@@ -4,38 +4,36 @@ import { EvolutionsComponent } from './evolutions';
 import { iconClass }           from '../utils/pokemon';
 import { setCurrentPokemon }   from '../actions/pokemon';
 
-export function EvolutionFamily ({ dex, family, setCurrentPokemon }) {
+const EvolutionColumn = ({ dex, evolutionStep, doesNotEvolve, setCurrentPokemon }) => (
```
robinjoseph08: Thanks for the refactor, but I think for now, we just want to keep it inline. There's not too much going on right now to warrant pulling it out (though it might be in the future if we add more functionality). Could you just add the `title` attribute to the 3 `<a>` tags? It would also simplify the diff considerably.

robinjoseph08:
@SimenB do you think you can make the changes mentioned? If you want, we can just implement this feature in another PR.

SimenB:
@robinjoseph08 Updated 😄

app/components/evolution-family.jsx
```diff
@@ -48,10 +48,6 @@ function mapStateToProps ({ currentDex, currentUser, users }) {
   return { dex: users[currentUser].dexesBySlug[currentDex] };
 }
 
-function mapDispatchToProps (dispatch) {
-  return {
-    setCurrentPokemon: (id) => dispatch(setCurrentPokemon(id))
-  };
-}
+const mapDispatchToProps = { setCurrentPokemon };
```
SimenB: Can't resist this small refactor

PR #255, opened 2017-02-20
chore(yarn): switch to yarn

### LINKED ISSUE #252: Switch from npm to yarn
- update each direct dependency
- remove the `npm-shrinkwrap.json` in favor of `yarn.lock`

robinjoseph08:
ꉂ(σ▰˃̶̀ꇴ˂̶́)σ

fixes #252 

finally moving off stoopid npm! from now on, we'll be using yarn for all things. for you, this means running `yarn start` instead of `npm start`, `yarn` instead of `npm i`, and [others](https://yarnpkg.com/en/docs/usage) if you so need it. but before you can do any of this, you need to install yarn by running `brew install yarn`. once you've got it, you just need to nuke your current `node_modules`, run `yarn` to regen it, and then make sure everything still looks good!


ami:
got this error:

image

robinjoseph08:
what version of node are you using? `node -v`

ami:
`v5.9.0`

robinjoseph08:
hmm I should be more specific about the version then. I'll update that! for now, run `nvm install 5.12` and then try again

PR #258, opened 2017-03-11
Removed random full stop

GiacomoLaw:


ami:
thanks for the contribution @GiacomoLaw! however this is actually [MissingNo.'s true name](http://bulbapedia.bulbagarden.net/wiki/MissingNo.).

GiacomoLaw:
Apologies- never knew!

PR #280, opened 2017-10-22
[WIP] refactor(slug): use slugify instead of slug

robinjoseph08:
\>^)))< ～～

switch the package we use for slugs cause the old one

- hasnt been updated in 2 years
- has a security vulnerability
- downloads unicode info from unicode.org on every install and if you do it too many times, unicode.org will rate limit you -_-
- it reduces `yarn` time by ~6 seconds which is good

i was working on the donation stuff but then i needed to upgrade some packages (pr for that coming soon) and then i got rate limited by unicode.org which led me here.

the functionality of this slugging module should stay the exact same tho!

### DIFF
--- a/app/actions/user.js
+++ b/app/actions/user.js
@@ -1,5 +1,5 @@
 import { push } from 'react-router-redux';
-import slug     from 'slug';
+import slugify  from 'slugify';
 
 import { Config }       from '../../config';
 import { API }          from '../utils/api';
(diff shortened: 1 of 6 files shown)


McInkay:
Had a quick look, and it looks like the dependency you currently used has had quite a lot of updates, and they don't require downloading the file anymore.

robinjoseph08:
@McInkay thanks for the heads up! It does look a lot better now. I'll close this PR and open a new one for updating `slug` :+1:

PR #290, opened 2017-12-31
feat(usum): ultra sun ultra moon support

ami:
˭̡̞(◞⁎˃ᆺ˂)◞*✰

- [x] update dex create form
- [x] update dex edit form
- [x] update register form
- [x] replace generation label with game label
- [x] replace all other instances of `generation`
- [x] nuke region filter
  - nuke `regionCheck`
  - nuke `region` from global state and any `mapStateToProps`
  - grep for `region`
- [x] add notification for blog post
- [x] style notification
- [x] add bitly url to notification
- [x] ensure the notification is on all appropriate pages
- [x] update home + ashketchum08 dexes

fixes https://github.com/pokedextracker/pokedextracker.com/issues/282


ami:
@robinjoseph08 i got it working for dex create and edit but when i try to apply the same thing to register it doesn't work... so can you take a look? maybe i'm just doing it wrong :3

also, when i create a regional us or um dex, #807 zeraora exists but #804-6 aren't there. is this a bug on the backend or is there something here we have to fix?

i tried doing the region filter nuking but it was difficult so can you do that? 😇

robinjoseph08:
@ami ill take a look at register and the region filter!

are you sure they dont exist at all? keep in mind that [they dont come in between 802 and 807](https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_by_Alola_Pok%C3%A9dex_number_(Ultra_Sun_and_Ultra_Moon))

app/components/dex-create.jsx
```diff
@@ -7,14 +7,29 @@ import slug          from 'slug';
...
+  componentWillMount () {
+    this.reset();
+  }
+
+  reset (props) {
+    const { listGames } = props || this.props;
+
+    listGames()
+    .then((games) => {
+      games.sort((a, b) => b.order - a.order);
```
robinjoseph08: hmm we shouldnt need to order on the front end, but i see why you did it. ill see if i can do that on the backend

ami:
@robinjoseph08 ooOoooooooo

app/actions/games.js
```diff
@@ -0,0 +1,21 @@
+import { Config }       from '../../config';
+import { API }          from '../utils/api';
+import { checkVersion } from './utils';
+
+export const SET_GAMES  = 'SET_GAMES';
```
robinjoseph08: nit: whitespace

also, can you rename this file to be `game.js`? the action files are all singular

app/actions/games.js
```diff
@@ -0,0 +1,21 @@
...
+
+export function listGames () {
+  return (dispatch) => {
+    dispatch(checkVersion());
+
+    return API.get(`${Config.API_HOST}/games`)
+    .then((games) => {
+      dispatch(setGames(games));
+      return games;
```
robinjoseph08: since we're setting `games` on the global state (with the previous line), we dont need to return them directly; the components will get it from the state, not from the result of this promise

app/actions/games.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const SET_GAMES  = 'SET_GAMES';
+
+export function listGames () {
+  return (dispatch) => {
+    dispatch(checkVersion());
+
+    return API.get(`${Config.API_HOST}/games`)
+    .then((games) => {
+      dispatch(setGames(games));
```
robinjoseph08: nit: once this is the only line, you can one-line it

```js
.then((games) => dispatch(setGames(games)));
```

app/components/dex-create.jsx
```diff
@@ -7,14 +7,29 @@ import slug          from 'slug';
...
+import { listGames }      from '../actions/games';
 
 const NATIONAL_ONLY_GAMES = ['x', 'y', 'omega_ruby', 'alpha_sapphire'];
 
 export class DexCreate extends Component {
 
   constructor (props) {
     super(props);
-    this.state = { error: null, game: 'sun', regional: false };
+    this.state = { error: null, game: 'ultra_moon', games: [], regional: false };
```
robinjoseph08: ive changed the order returned from api to be

```
us
um
sun
moon
or
as
x
y
```

so this should be `game: 'ultra_sun'`

robinjoseph08: we wont be saving `games` in the local component state (i.e. `this.state = {}` and `this.setState({})`). instead, the action will be saving it in the global state. and instead of pulling it from `this.state`, we'll be pulling it from `this.props` because `mapStateToProps` will pull it form the global state and set it as the components `props`

app/components/dex-create.jsx
robinjoseph08: this has been deployed to api, so you can remove all sorting lines

app/components/dex-create.jsx
```diff
@@ -7,14 +7,29 @@ import slug          from 'slug';
...
+  }
+
+  componentWillMount () {
+    this.reset();
+  }
+
+  reset (props) {
+    const { listGames } = props || this.props;
+
+    listGames()
```
robinjoseph08: instead of calling this here, lets call it in the `profile` component. and since its setting `games` on the global state and wont be returning it, we can just call it and forget about it e.g.

```js
listGames();
```

app/components/dex-create.jsx
```diff
@@ -66,7 +81,7 @@ export class DexCreate extends Component {
 
   render () {
     const { isOpen, session } = this.props;
-    const { error, game, regional, url } = this.state;
+    const { error, game, games, regional, url } = this.state;
```
robinjoseph08: instead of pulling `games` from `this.state`, lets pull it from `this.props`

app/components/dex-create.jsx
```diff
@@ -66,7 +81,7 @@ export class DexCreate extends Component {
...
```
robinjoseph08: the fact that the whole component is being rendered when now, there are parts that need to be loaded async (i.e. the games), is proving to be problematic. cause when this `render` function is first called, `games` is `undefined` because the api call hasnt finished yet. so youll get the error `Cannot read property 'map' of undefined`. so we need to prevent the rest of the component from rendering unnecessarily. i think we can just accomplish this with

```js
if (!isOpen || !games) {
  return null;
}
```

this should be fine for us cause its just a precaution since were loading games from the profile and itll take some time before someone clicks the edit of create button. the alternative is to have a `loading` variable and a loading state and stuff. kinda think its overkill here but ill leave it to your discretion. i think we might need it for the register tho since the results of the fact that the games arent there will be apparent, again, ill leave it to you

app/components/dex-create.jsx
```diff
@@ -135,6 +149,7 @@ function mapStateToProps ({ session }) {
 function mapDispatchToProps (dispatch) {
```
robinjoseph08: cant comment on the line, but add `games` to the args and the return value of `mapStateToProps`

app/components/dex-edit.jsx
```diff
@@ -20,6 +21,7 @@ export class DexEdit extends Component {
     this.state = {
       error: null,
       game: props.dex.game.id,
+      games: [],
```
robinjoseph08: most of the same comments from dex create

app/components/dex-edit.jsx
```diff
@@ -59,11 +75,17 @@ export class DexEdit extends Component {
...
 
-  get showGenerationWarning () {
+  get showGameWarning () {
     const { dex } = this.props;
     const { game } = this.state;
 
-    return game === 'omega_ruby' && dex.game.id === 'sun';
+    if ((dex.game.game_family.id === 'sun_moon') && NATIONAL_ONLY_GAMES.indexOf(game) > -1) {
+      return true;
+    } else if ((dex.game.game_family.id === 'ultra_sun_ultra_moon') && game !== 'ultra_sun' && game !== 'ultra_moon') {
```
robinjoseph08: blech. ill think of way to make this better

app/components/info-locations.jsx
```diff
@@ -44,7 +52,7 @@ export function InfoLocationsComponent ({ gameFamily, pokemon, regional }) {
     </div>
   );
 
-  const gen7 = (
+  const sun_moon = (
```
robinjoseph08: nit: camelCase

app/components/info-locations.jsx
```diff
@@ -57,10 +65,24 @@ export function InfoLocationsComponent ({ gameFamily, pokemon, regional }) {
     </div>
   );
 
+  const us_um = (
```
robinjoseph08: nit: camelCase

app/components/info-locations.jsx
```diff
@@ -57,10 +65,24 @@ export function InfoLocationsComponent ({ gameFamily, pokemon, regional }) {
...
+      </ul>
+    </div>
+  );
+
   return (
     <div className="info-locations">
-      {gameFamily.generation === 7 ? gen7 : null}
-      {regional ? null : gen6}
+      {gameFamily.id === 'ultra_sun_ultra_moon' ? us_um : null}
+      {gameFamily.id === 'sun_moon' || gameFamily.id === 'ultra_sun_ultra_moon' && !regional ? sun_moon : null}
```
robinjoseph08: blech. will rethink. in the meantime, you should wrap the `||`s in parens

app/components/info.jsx
```diff
@@ -37,7 +37,7 @@ export class Info extends Component {
 
   render () {
     const { dex, pokemon, showInfo } = this.props;
-    const serebiiPath = dex.game.game_family.generation === 6 ? 'pokedex-xy' : 'pokedex-sm';
+    const serebiiPath = dex.game.game_family.id === 'x_y' || dex.game.game_family.id === 'omega_ruby_alpha_sapphire' ? 'pokedex-xy' : 'pokedex-sm';
```
robinjoseph08: using `generation` is fine here 👍

app/components/register.jsx
```diff
@@ -23,12 +24,22 @@ export class Register extends Component {
   }
 
   componentWillMount () {
-    const { checkVersion, redirectToProfile, session } = this.props;
+    this.reset();
+  }
+
+  reset (props) {
```
robinjoseph08: i dont think we need to break this out into `reset`

app/components/register.jsx
```diff
@@ -23,12 +24,22 @@ export class Register extends Component {
...
       redirectToProfile(session.username);
     }
 
+    listGames()
+    .then((games) => {
+      games.sort((a, b) => b.order - a.order);
+      this.setState({ games });
+    });
+
     checkVersion();
```
robinjoseph08: since `listGames` calls `checkVersion`, you can remove this call here

app/components/register.jsx
```diff
@@ -23,12 +24,22 @@ export class Register extends Component {
...
+  }
+
+  reset (props) {
+    const { checkVersion, listGames, redirectToProfile, session } = props || this.props;
 
     if (session) {
       redirectToProfile(session.username);
     }
 
+    listGames()
```
robinjoseph08: similar stuff to the other 2 components

app/components/dex-create.jsx
ami: makes sense - i think it's fine here but we should add one to register.

app/components/dex-edit.jsx
ami: ya it is confoosing lol.

app/components/dex-create.jsx
ami: ok nvm it's loading very quickly, so not gonna add a loading state to register either 😂

ami:
@robinjoseph08 updated! and register works now too yeyyy

PR #303, opened 2018-05-13
test(circle): switch from travis to circle

robinjoseph08:
ヘ(￣ω￣ヘ)

- this makes it so we only build once instead of the twice that travis did
- instead of averaging ~3m per build, its averaging ~30s per build which is nice

since its just infra-y things, im just gonna merge!


ami:
wow what if i wanted to review this

robinjoseph08:
ill rmbr to assign all infra prs to you in the future 👍

PR #312, opened 2018-12-22
feat(locations): use pokemon.locations to simplify logic

robinjoseph08:
( ¨̮ )

this is using the new `locations` array instead of the custom logic we had before. the main changes involve showing all locations for national dexes for the selected game's family's generation and below. so for the common case, the main change will be that it shows the usum locations for sumo national dexes.

i also added some health check stuff but that dont matter


app/components/info-locations.jsx
```diff
@@ -1,89 +1,16 @@
...
-      </ul>
-      <h3>Pokémon Ultra Moon</h3>
-      <ul>
-        {ultraMoonLocations}
-      </ul>
-    </div>
-  );
-
-  // TODO: refactor this logic to more scalable and easier to understand
+export function InfoLocationsComponent ({ pokemon }) {
```
ami: w o w 🙏

PR #314, opened 2018-12-25
feat(games): use regional_support and national_support

### LINKED ISSUE #311: Editing dex's available Pokemon doesn't update until page is refreshed
When creating a regional dex, and then editing it to be a nation dex, the dex itself still retains only the region pokemon.

robinjoseph08:
°₊·ˈ∗(( ॣ>̶᷇ᗢ<̶᷆ ॣ))∗ˈ‧

- instead of relying on `NATIONAL_ONLY_GAMES`, rely on db value of `regional_support`
  - i also made the db have `regional_support = true` for xy, oras, sumo, and usum so we'll support regional gen 7 dexes as well
- rely on `national_support` to potentially not allow national dexes for a game. this isnt used now, but will be for lets go
- the default value for "new dex" is always the latest game
- change all references to specific games (e.g. the word alolan in various places)
- fixes #311

### DIFF
--- a/app/actions/game.js
+++ b/app/actions/game.js
@@ -9,7 +9,10 @@ export function listGames () {
     dispatch(checkVersion());
 
     return API.get(`${Config.API_HOST}/games`)
-    .then((games) => dispatch(setGames(games)));
+    .then((games) => {
+      dispatch(setGames(games));
+      return games;
+    });
   };
 }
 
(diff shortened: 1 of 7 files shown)


robinjoseph08:
nice catch! fun fact, it was actually the same issue so fixing one fixed the other lol

PR #315, opened 2019-01-03
feat(lets-go): add new spritesheet/css

robinjoseph08:
ε=ε=ε=┏(ﾟロﾟ;)┛

update the spritesheet and the css to have meltan and melmetal. the 50px really put a damper on things, but i think its good! if you go to `/u/hiami/regional-lets-go-pikachu` while running this locally, then you should be able to see meltan and melmetal too


ami:
@robinjoseph08 yay glad the spritesheet worked out! the only issue i saw was that the info panel looked a bit crammed for melmetal, so i added a commit that has some styles just for melmetal. lmk if you think it's too hacky... 

here's what it looks like after:
image

robinjoseph08:
blarg... the custom styles is a bit hacky... in the scss generation script, we're already doing some custom stuff for "full" sprites eg 50px https://github.com/pokedextracker/pokesprite/pull/4/files is there anyway we can add these styles there so that if in the future, more full sprites come along, it won't get even more bespoke?

ami:
yep if you add a class like `width-50` or something i could use that instead?

robinjoseph08:
sorry, not sure i follow. right now, the script generates that scss file that contains `.pkicon.pkicon-809` with `background-position`. what specifically (classes and/or styles) do you want me to add?

ami:
ooh.. right.... blegh let's just leave it like this then....?

robinjoseph08:
blargggg ya lets leave it as is. idk if you wanna leave any comments to help future ami when adding new styles for new pokemon, but im gonna merge this and you can add that if you want!

PR #316, opened 2019-01-03
feat(lets-go): support switch friend codes

ami:
(>^_^)><(^o^<)

^ they're FRIENDS!!

- updates references from `friend_code` to `friend_code_3ds`
- adds ability to add/edit `friend_code_switch` on registration and account edit
- shows both friend codes on profile and dexes


app/components/friend-code.jsx
```diff
@@ -8,7 +8,7 @@ export function FriendCode ({ session, user }) {
 
   return (
     <h2>
-      FC: <span className={user.friend_code ? '' : 'fc-missing'}>{user.friend_code || 'XXXX-XXXX-XXXX'}</span>
+      <b>3DS FC</b>: <span className={user.friend_code_3ds ? '' : 'fc-missing'}>{user.friend_code_3ds || '0000-0000-0000'}</span>, <b>Switch FC</b>: <span className={user.friend_code_switch ? '' : 'fc-missing'}>{user.friend_code_switch || 'SW-0000-0000-0000'}</span>
```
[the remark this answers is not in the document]
ami: oops prob an oversight. i'll switch it back to Xs!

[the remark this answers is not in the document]
ami: i originally wanted to do stacked, but then it seemed weird to have the edit icon next to both, when they both take you to the account edit page. what do you think? could go either way!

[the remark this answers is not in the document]
robinjoseph08: hmm ya it is a bit odd... but i think i prefer that over the side by side. is there any way to make it look good with stacked and one edit button? i think the answer is no, but maybe you can work some magic :3 if not, two buttons is fine lol

ami:
@robinjoseph08 your changes look good! also updated fcs to be stacked:

image

app/components/friend-code.jsx
```diff
@@ -1,16 +1,28 @@
...
+      <Link to="/account" onClick={() => ReactGA.event({ action: 'click edit friend code', category: 'User' })}><i className="fa fa-pencil" /></Link>
+    );
+  }
+
   return (
-    <h2>
-      FC: <span className={user.friend_code ? '' : 'fc-missing'}>{user.friend_code || 'XXXX-XXXX-XXXX'}</span>
-      {ownPage ? <Link to="/account" onClick={() => ReactGA.event({ action: 'click edit friend code', category: 'User' })}><i className="fa fa-pencil" /></Link> : null}
-    </h2>
+    <div>
```
ami: would be nice to upgrade to v16 of react at some point so we can use [fragments](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) 😄

robinjoseph08: TIL

app/components/friend-code.jsx
```diff
@@ -1,16 +1,28 @@
-import { Link }    from 'react-router';
-import { connect } from 'react-redux';
+import { Link }     from 'react-router';
+import { connect }  from 'react-redux';
```
robinjoseph08: super nit: these spaces arent necessary

app/components/friend-code.jsx
ami: we use them in lens!!!

PR #317, opened 2019-01-03
feat(lets-go): launch lgpe support

### LINKED ISSUE #313: Add support for Let's Go
- [x] convert to using `locations` on the frontend
- [ ] rely on `regional_support` and `national_support` instead of a hard-coded array
- [ ] add unpublished games/game family to the db
- [ ] add new pokemon to db
- [ ] add locations to db
- [ ] add new spritesheet and css
- [ ] update homepage
- [ ] update games/game family to be published

ami:
from the ticket, this should encompass:
- add notification (the blog post link isn't live yet, so it's broken for now)
- update homepage
- support candy evolution in ui (looks like this works as is, so i don't think any code changes are required) 

fixes #313


app/components/home.jsx
```diff
@@ -41,8 +41,18 @@ export class Home extends Component {
             </div>
 
             <div className="sub">
-              <h2>A tool for tracking your Living Dex progress for Pokémon Ultra Sun &amp; Ultra Moon (both Gen 7 <Link className="link" alt="Sample Ultra Sun Regional Living Dex" to="/u/ashketchum10/ultra-sun-regional-living-dex">Alola Dex</Link> and <Link className="link" alt="Sample Ultra Sun National Living Dex" to="/u/ashketchum10/ultra-sun-national-living-dex">National Dex</Link>) or <Link className="link" alt="Sample Omega Ruby Living Dex" to="/u/ashketchum10/omega-ruby-living-dex">Gen 6</Link>. Easily toggle between and track your captured Pokémon, find the locations of those left to be captured, and share a public link with others to see how you can help each other out. Manage all your dexes on one <Link className="link" alt="Sample Profile" to="/u/ashketchum10">profile</Link>, and even track a <Link className="link" alt="Sample Shiny Living Dex" to="/u/ashketchum10/shinies">shiny living dex</Link>!</h2>
-
+              <h2>A tool for tracking your Living Dex progress! We currently support:</h2>
+              <ul>
```
ami: i ended up completely reworking this to a bulleted list of what we support. lmk if you don't like it...

also, do you remember why this was an `h2` but the final paragraph is a `p` (yet they are styled the same)? was it for SEO?

robinjoseph08: took forever to track this down, but ya looks like its for seo lol https://github.com/pokedextracker/pokedextracker.com/issues/57#issuecomment-207697468

ami: ok.. so should we keep it? and does that mean by putting the dexes in `ul/li` that it won't go towards SEO?

robinjoseph08: uhh its not that it wont go towards seo, its that we wanted the things in the h2 to be given more weight than everything else (e.g. like the made in san francisco stuff) so we made them more important. tbh idk how ul/li ranks in terms of seo, but i dont think its as high as h2. so i think it would be _better_ to have those pieces of text within the h2, if at all possible? is it weird to have the ul/li in the h2?

ami: it's not semantic to put `ul/li` in `h2` because `h2` is a block element. buuttt i can wrap the text that's in the `li` with `h2`. so i'll do that!

ami:
@robinjoseph08 rebased + updated based on comments! fixed some other responsiveness stuff too.

PR #331, opened 2019-06-18
Update node-sass for macOS Mojave

briaguya:
Fixes "Node Sass does not yet support your current environment" error when running on Mojave (macOS 10.14.5)


robinjoseph08:
@briaguya apologies for the super long delay here. I've upgraded both bode and node-sass in #364, so this should now be working!

PR #339, opened 2019-12-08
feat(swsh): modify box logic for sword and shield

robinjoseph08:
ミ(・・)ミ

- don't make boxes exclusive to national dexes
- use `uniqBy` to make unique map keys cause we were getting an error because there were some duplicate evolutions (e.g. for the hitmons, they all had a `breed` evo, and they were all dupes)
- change the local endpoint to just be localhost cause its easier


robinjoseph08:
issues:

1. hm im not entirely sure what you mean? it looks normal to me? can you send a ss?
2. updated!
3. updated!
4. updated!

questions:

1. it requires an api change, but i can change the ordering. that said, i felt like this ordering made a bit more sense since non-galarian forms are real differences in the pokemon, whereas gigantamaxes are still just normal pokemon with a special trait. so in the actual box in the game, they will look different. but that was just my thought process.
2. there is precedent for multi evolution methods that changed (e.g. leafeon and glaceon), but tbh, i see this as a stopgap. long term solution, especially if we plan on supporting the older gen's, is to do a similar thing like we did for boxes and affiliate an evolution with a specific game family. but i dont think that should block this launch tho. i've added it to the "after release" section of the epic issue
3. yaa i noticed that too. i dont think its a huge deal, but i would be okay with hardcoding it just for them. tho idk how we would structure it if we even wanted to do it "the right way"
4. answered in slack!

ami:
ignore the font size, i zoomed way out to show the issue, but basically it looks like the alolan forms keep getting put in their own boxes:

image

it's weird that you don't see it... o.o hmm.

as for mr.rime, i think the "right" way to do it is that you would know on the backend which evolutions branch from what, and you could insert like a fake blank "div" that takes up space based on some logic. maybe? for now i think i'll just add in a manual thing to push it down for just mr. rime haha. i can do that in my PR.

ok to everything else! 👍 

PR looks good other than the box issue for LGPE, SM, and USUM regional dexes (for me)..

robinjoseph08:
i figured it out. turns out i was running the wrong branch before lul. to fix it, i had to make an api change, but its been done and deployed, so it should be fixed now. i also found another bug with boxes as well, but i just merged the fix and will be deploying it shortly

robinjoseph08:
lgtm!

PR #347, opened 2020-01-18
feat(swsh): use shiny sprites

robinjoseph08:
(۶* ‘ꆚ’)۶”

use the swsh shiny sprites provided by KirinLP! i also renamed shellos east and west so we dont need the hack anymore!


robinjoseph08:
@ami good catch! looks like i only trimmed some of the new sprites :sweat_smile: they should all be good now!

PR #352, opened 2020-02-09
feat(swsh): national dex support

### LINKED ISSUE #349: Add National Dex for Sword and Shield

ami:
└（★ｏ★）┐

fixes #349

### DIFF
--- a/app/components/home.jsx
+++ b/app/components/home.jsx
@@ -43,12 +43,11 @@ export class Home extends Component {
             <div className="sub">
               <h2>A tool for tracking your Living Dex progress! We currently support:</h2>
               <ul>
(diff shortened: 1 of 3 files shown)


app/components/home.jsx
```diff
@@ -43,12 +43,11 @@ export class Home extends Component {
             <div className="sub">
               <h2>A tool for tracking your Living Dex progress! We currently support:</h2>
               <ul>
+                <li><h2><Link className="link" alt="Sample Generation 8 National Living Dex" to="/u/ashketchum10/sword-national-living-dex">Generation 8 National Dex</Link> (and Generations 7 + 6)</h2></li>
```
ami: i decided to combine these cuz the list was getting long, thoughts?

robinjoseph08: works for me! thoughts on making it `6 + 7` vs `7 + 6`? it reads a bit better that way

PR #358, opened 2020-02-16
feat(night-mode): enable as an option

### LINKED ISSUE #234: Add night/dark mode
to make using the app at night easier on the eyes

ami:
（／．＼）

- adds night mode local storage
- adds night mode toggle to nav dropdown (the only downside to this is you can only toggle if you're signed in. thoughts?)
- adds styling for night mode.. open to suggestions for the palette i chose! i'm not sure if the yellow for captured is still too bright. also thinking about making the main top nav dark blue.

Screen Shot 2020-02-16 at 3 43 48 PM
Screen Shot 2020-02-16 at 3 44 06 PM

fixes #234


app/components/dex-create.jsx
```diff
@@ -84,15 +84,15 @@ export class DexCreate extends Component {
   }
 
   render () {
-    const { games, gamesById, isOpen, session } = this.props;
+    const { games, gamesById, isOpen, session, nightMode } = this.props;
```
ami: looks like the modals get added outside of root, so i had to maually add this to both `dex-create` and `dex-edit`. open to doing something else if you see a better option!

app/components/app.jsx
```diff
@@ -1,3 +1,4 @@
+import { connect }                       from 'react-redux';
```
ami: when i toggle night mode it posts this error in the console:

`Warning: [react-router] You cannot change <Router routes>; it will be ignored`

i think due to this file? but i'm not sure.

McInkay: I think this is because when you change nightmode, it'll rewrite the whole app including the router, but router doesn't support changing the routes after the initial time. 

It might be better to have the `night-mode` class applied to the individual components rather than the overall one. You could create a small wrapper component to wrap each of the `HomeComponent`, `LoginComponent` etc that would do the nightmode so you aren't duplicating the code.

ami:
@robinjoseph08 i originally started with a mostly black palette but still wanted to keep some blues/yellows for brand. can you push up what you have and i’ll modify it a bit using the guidelines you posted to see how i can put some more blues/yellows in that still fit within the guidelines? :D

robinjoseph08:
yup, pushed! i also pushed up the router change since i had already done it. also my style changes are all over the place, soooo :see_no_evil:

ami:
@robinjoseph08 updated! moved the toggle into the top nav + updated lotsa colors.

app/components/nav.jsx
```diff
@@ -75,6 +86,10 @@ export class Nav extends Component {
     return (
       <nav>
         <Link to="/">Pokédex Tracker</Link>
+        <a onClick={this.toggleNightMode} className="tooltip tooltip-below">
+          <i className="fa fa-adjust" />
+          <span className="tooltip-text">Night Mode {nightMode ? 'Off' : 'On'}</span>
+        </a>
```
ami: would be nice if we could move these 3 elements (home link, night mode toggle, and patreon link) to a shared var so we don't have to repeat them twice -- but i can't do that because it has to be enclosed as 1 element. wrapping it in a div messes up the styling, and we don't have fragments. thoughts?

robinjoseph08: updated to use a fragment!

robinjoseph08:
the tooltip looks fine! but it looks like it gets covered by the info pane if its open, so we should fix that

ami:
dunn

robinjoseph08:
@ami moon icon added!

ami:
@robinjoseph08 added the toggle and some minor style fixes. last look?? :D

PR #359, opened 2020-02-19
Hide captured pokemon when "Hide pokemon" is checked

McInkay:
I added a small feature to allow people to hide captured pokemon. 

I put it under the search bar, but not sure if you can think of a better place for it to go.

A few things that probably need to be done:
- [ ] Fix styling issues, it looks kinda ugly, probably need to make it nicer
- [ ] Probably save the setting in local storage - it currently gets cleared, just like the seearch query, but this probably makes more sense to persist

In terms of how this works, I made a few design decisions. Feel free to tell me I'm wrong with any one them.

Firstly, the [default view](https://files.yamanickill.com/gu81AvsK.png) looks exactly the same, just with the "hide completed" checkbox appearing under the search box.

When you check this, [those pokemon are hidden from view](https://files.yamanickill.com/pJFXLIiH.png). We just don't render them.

You'll notice that it doesn't show all 30 spaces in this screenshot. It would defeat the purpose of this feature to do so, as you'll jsut end up with boxes of 30 empty, which makes you scroll as much as you would before. So I've basically modulo-ed the boxes to 6. So if there are 28 in it, you'll get 30 spaces, but if there's 1, you'll get 6, 8, you'll get 12, etc.

This also lowers the whitespace for the [alola/galar sections](https://files.yamanickill.com/RzyRI-RN.png), they take up fewer rows now.

Next, [when you have a box fully done](https://files.yamanickill.com/qFqydPO-.png), it'll completely hide from view. I thought this made most sense as possibly seeing a box with no pokemon would be a bit confusing, and if you are selecting "Hide Completed" it'll be to save space and mental calculations on looking through the list. So this made most sense to me.

Thanks for the website, I really like it, and thought I'd add the code for this feature that I'd really like. Happy to modify things as you want. Let me know if you want this feature or not. As shown above, there are a couple of things I need to do before this would be ready for merge anyway, but thought I'd get it up now before I spend too much time pouring over CSS.

### DIFF
--- a/app/actions/search.js
+++ b/app/actions/search.js
@@ -1,5 +1,10 @@
 export const SET_QUERY = 'SET_QUERY';
+export const TOGGLE_FILTER_COMPLETED = 'TOGGLE_FILTER_COMPLETED';
 
 export function setQuery (query) {
   return { type: SET_QUERY, query };
+}
+
+export function toggleFilterCompleted () {
+  return { type: TOGGLE_FILTER_COMPLETED };
 }
--- a/app/components/box.jsx
+++ b/app/components/box.jsx
@@ -5,10 +5,22 @@ import { PokemonComponent }       from './pokemon';
 import { deferComponentRender }   from '../utils/defer-component-render';
 import { padding }                from '../utils/formatting';
 
-export const BOX_SIZE = 30;
+export const STANDARD_BOX_SIZE = 30;
 
-export function Box ({ captures, dex }) {
-  const empties = Array.from({ length: BOX_SIZE - captures.length }).map((_, i) => i);
+export function Box ({ captures, dex, hideCompleted }) {
+  const showPokemon = showPokemonCurried(hideCompleted);
+  const pokemonToShow = captures.filter((capture) => showPokemon(capture));
+  let BOX_SIZE;
+  // We want to have every box in a line filled, but no need to add extra lines if we don't need to
+  if (pokemonToShow.length % 6 === 0) {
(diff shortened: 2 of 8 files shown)


ami:
hey @McInkay, thanks for the contribution! if you check out our [roadmap](https://github.com/pokedextracker/pokedextracker.com/projects/1), we’re actually going to be working on this feature soon, but we were planning on implementing a bit differently than the route you’ve chosen to go with the boxes. as such, i’m going to close this PR for now but we may use it for some code inspiration when we’re implementing our version :slightly_smiling_face:

in the future if there are some changes you want to make, it’s probably a good idea to comment on the issue for it to see how it aligns with our plans for the site and/or make sure it isn’t something we’re already in the middle of working on. if it isn’t, and there’s room for an external contribution, then a comment will helps us guide you on what we were thinking for the functionality and ui/ux. this will help us both out to make sure were all on the same page before any time is spent!

McInkay:
I mean, I'm happy to make the changes if there's something in how I did it you disagreed with. This was my first version of the feature, and was just the way I thought it should work. Happy to polish it up and change some of the logic if you want it done differently.

To be honest, it was quicker to implement this than to comment and then wait to see what happened. I didn't polish it up deliberately so I didn't spend too long working on it to find out you didn't want it that way, that's fine.

I did notice it was on your roadmap, but as far as I could tell it wasn't currently in progress, so I thought I could help out. There's no point in duplicating the work that is already done in it, unless this is hugely different from how you want it done.

PR #364, opened 2020-02-29
chore(upgrade): upgrade react to v16 (and a bunch of other deps)

robinjoseph08:
<゜)))彡

update the `slug` package. this is an update of #280 now that `slug` has updated a lot. it seemed like the only "breaking changes" were the additions of arabic and farsi slugification, and by checking in the db, it didn't seem like there were any usernames with those characters, but it's a bit tough to say for certain.


robinjoseph08:
okay, i kinda went a bit crazy and updated pretty much everything... its now using the latest version of all our deps (and i was able to remove a few unused ones).

i also converted everything to functional components and we're now using [hooks](https://reactjs.org/docs/hooks-intro.html). this means that all of our components can be functional ones (we never have to use `class *Component extends Component`) and theres a lot of cool things we can do with hooks. and its generally less code. and now we have fragments too! im excited about it :3 

tbh you dont have to look through all the code, but just run through all the flows and make sure no errors show up in the console. i think i checked everything myself several times, but you're good at finding things ive missed

and dont worry too much about learning all the new stuff. pattern matching will get you far and anything else ill help you with when we hit issues in the future :+1: 

tho ironically, the only thing i didnt update is font awesome yet... but ill be doing that soon too!

PR #367, opened 2020-03-06
feat(info): Add filter to show only needed Pokemon

doeiqts:


doeiqts:
I believe this should satisfy https://github.com/pokedextracker/pokedextracker.com/issues/136

While I tried to be as careful as I could, I've never messed with React-Redux (do I even have that right?), so if there's issues where I've just done something ridiculous let me know and I'll try and fix it.

I did test locally and it seemed to be doing what I wanted it to finally. Hopefully, this is helpful. I've been interested in this feature for a while, so I think it'll be nice to have.

doeiqts:
Is there anything else I need to do or can help with to get this merged? Just let me know. Thanks.

robinjoseph08:
Hey @doeiqts, thanks for the contribution! Unfortunately, it's coming at a slightly challenging time because we're reworking some of the architecture of this repo. The most notable one being the upgrade from React v15 to v16 and utilizing [hooks](https://reactjs.org/docs/hooks-intro.html) pretty heavily. That just got merged in pretty recently (#364), and that's the cause for the merge conflicts.

But we wholeheartedly agree that this is a helpful feature that a lot of people would enjoy! We've actually had our eye on it ourselves for a bit, but it's unfortunately tougher to implement for two main reasons:

1. You may have noticed it yourself while testing locally, but when dealing with filtering the captures, there seems to be a performance hiccup when enabling the filter. The browser freezes for a second as it attempts to rerender all of the `Pokemon` components. We actually saw a similar lag occur when we first implemented the search bar, and it took a bit of time to figure out the best way to overcome it so that we can avoid the noticeable pause. I'd assume it'd be a bit more trail-and-error this time around as well.
2. We've been thinking about the exact UX of this feature that makes the most sense, specifically around what happens when you mark a Pokemon as captured. The obvious way is to have that Pokemon disappear immediately, which is what happens in your PR. But this is less than ideal in the scenario of accidental markings because then you have to remove the filter and find the mis-marked Pokemon again to undo the action. We think there's a better way to approach it, but we're not 100% sure what that is yet. Two examples includes: 1) having a delay on when the Pokemon disappears, so you can quickly undo mistakes and 2) keeping the Pokemon you marked around, but then having an explicit "refresh" button that clears out the newly-marked ones.

Because of these things, we've decided that this is a feature that we're going to be implementing ourselves. And it's definitely on the horizon for us! After we finish up dark mode (#358), this is going to be the next feature we work on, so you should see it very soon.

All that said, we still appreciate your effort! When we do end up tackling this, we'll probably use some of the code that you've provided here to kickstart us in the right direction.

Thanks again, and we hope you understand!

PR #368, opened 2020-03-10
feat(icons): upgrade fontawesome from v4 to v5 with svgs

robinjoseph08:
୧༼✿ ͡◕ д ◕͡ ༽୨

- upgrade from fontawesome v4 to v5 so we get access to all the new icons
- switch from the webfont/`<i>` version to the svg version. this gives us a more crisp icon and no need for network calls anymore cause its all embedded svg. but since its a different html element, it required a decent amount of refactoring of styles to get things to look the same.

while i got most of the positioning correct, some of the new icons just look different, but there's not much we can do about that. i think generally everything looks fine. the only noticeable one is the `mars` and `venus` icons for the nidrorans. its a bit more bold now. but again, idk if there's anything we can do about it per se


ami:
LGTM! i'm ok with not making them thinner since it seems that's pro only. i guess the only alternative is to use a unicode character?

i think we now have access to use [circle-notch](https://fontawesome.com/icons/circle-notch?style=solid) as a spinner instead of the dotted one... wanna change to that one? i think it's prettier.

also i noticed on the save account page (as i was looking into the spinners) that the arrow right doesn't go away when it toggles to `isLoading`. happens in prod too. but the code looks right to me at a brief glance so not sure what's going on. weird! if you could fix that too that'd be cool, even tho it's unrelated.

robinjoseph08:
yup, the only alternative is unicode for them

updated to use circle-notch and fixed the weird arrow bug!

app/components/account.jsx
```diff
@@ -164,8 +164,9 @@ export function Account () {
             <FontAwesomeIcon icon={faChevronDown} />
           </div>
           <button className="btn btn-blue" type="submit">
-            <span className={isLoading ? 'hidden' : ''}>Save <FontAwesomeIcon icon={faLongArrowAltRight} /></span>
-            {isLoading ? <span className="spinner"><FontAwesomeIcon icon={faSpinner} spin /></span> : null}
+            {/* The double check for isLoading is necessary because there is a slight delay when applying visibility: hidden onto the icon. */}
```
ami: weird o.o

PR #369, opened 2020-03-21
feat(filter): add hide caught filter

### LINKED ISSUE #136: Have a view of only uncaught pokemon

robinjoseph08:
(❍ᴥ❍ʋ)

fixes #136 

- refactored search query to not use redux for simplicity and making the flow of data explicit
- add the hide caught filter
  - i put the checkbox in the `SearchBar` component, but feel free to make a new component or structure it however you like! i'm sure you'll be able to pattern match for it!
  - i tried addressing some of the performance stuff, but idk how its turning out cause my laptop is trash and everything is laggy anyway. lmk how it feels when you use it!


ami:
@robinjoseph08 rebased + added styles, want to take a look?

ezgif com-gif-maker

ami:
one thing i noticed, if you search with the filter, the pokemon don't show up (makes sense). however, if you press the "X" button or press "Clear your search?" the filter doesn't get reset. do you think it should? or should the X/link only be for the search? if we do allow it to also clear the filters, we can probably change the text to "Clear your search and/or filters?"

image

this can also happen if you have a 100% dex and you apply the filter.

Azekthi:
The "X" and "Clear your search" should _not_ disable the filter, since those links do not imply disabling the filter. Instead, the "no results" with the active filter should say something like this (modeled after Gmail):

> No results in uncaught Pokémon. [Show all Pokémon], or [clear your search]?

The brackets indicate links. Clicking "Show all Pokemon" should disable the filter but keep the search entry, while clicking "clear your search" should do the opposite, clear the search entry but keep the filter.

If the filter is enabled but there's no search entry, it should say something like:

> No uncaught Pokémon. [Show all Pokémon?]

If the filter is disabled and no results are found, it should maintain the current behavior i.e.

> No results. [Clear your search?]

robinjoseph08:
yup, i agree with @Azekthi's expectations. i think i might change some of the verbiage (e.g. "Show all Pokemon" can be a bit unclear on what it would do, so being more clear that it's just removing the caught filter would be better), but i'll implement the different cases!

Azekthi:
That change to make the first message say "Include caught Pokémon" is good 👍

ami:
@robinjoseph08 rebased - LGTM!

PR #377, opened 2020-05-22
fix(night-mode): fix reload background in night mode

DuBistKomisch:
got blinded by this one today 😛

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -994,6 +994,10 @@ select {
       color: $night-mode-error-dark
     }
   }
+
+  .reload {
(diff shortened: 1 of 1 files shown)


app/styles/main.scss
```diff
@@ -994,6 +994,10 @@ select {
       color: $night-mode-error-dark
     }
   }
+
+  .reload {
+    background-color: $night-mode-primary-dark;
```
ami, line 999: could we use `$night-mode-16dp` instead?

DuBistKomisch: I was trying to stick with `primary` but that does look better:

image

image

PR #378, opened 2020-05-22
fix(search): allow searching anywhere in the name

DuBistKomisch:
not sure if this was intentional but I thought I was going crazy when this didn't work


robinjoseph08:
Thanks for the contribution, @DuBistKomisch!

When we added this functionality, we were intentional about having it match from only the start of the name. In the years that we've had it, we haven't heard any complaints about the way that it filters, and so we're a bit wary to change it. However, if we do hear more feedback that it should filter from any part of the name, then we'll consider adding this change. But for now, we'll be closing it. Hope you understand!

DuBistKomisch:
no worries, fair enough, just struck me as a bit weird

PR #435, opened 2022-02-04
feat(box): add intemediate state for each Pokemon cell

### LINKED ISSUE #433: Pending API call view
Hey folks, I have noticed that when creating a lot of operations at once such as filling out a dex for the first time, many calls take a bit of time to complete. This is completely understandable, but the UI only updates once the call has been completed. When this takes several seconds, it may cause the UI to look unresponsive or make me think that I didn't actually click it.

It would be nice to add some pending state, like a yellow-white gradient or some other pattern to indicate that a click was registered and now we're waiting.

Fleker:
While an API call is in-progress, add a 'pending' class which is a gradient between two colors.
This provides the user with an acknowledgement of the action while API calls may be pending.

Fixes #433

An example of how this looks is the Pachirisu shown in the screenshot below.
image

### DIFF
--- a/app/actions/capture.js
+++ b/app/actions/capture.js
@@ -2,12 +2,15 @@ import { API }          from '../utils/api';
 import { Config }       from '../../config';
 import { checkVersion } from './utils';
 
+export const MARK_PENDING = 'MARK_PENDING';
 export const MARK_CAPTURED = 'MARK_CAPTURED';
 export const SET_CAPTURES = 'SET_CAPTURES';
+
 
 export function createCaptures ({ payload, slug, username }) {
   return (dispatch) => {
     dispatch(checkVersion());
+    dispatch(markPending(payload.pokemon, slug, username));
 
     return API.post(`${Config.API_HOST}/captures`, payload)
       .then(() => dispatch(markCaptured(true, payload.pokemon, slug, username)));
@@ -17,6 +20,7 @@ export function createCaptures ({ payload, slug, username }) {
 export function deleteCaptures ({ payload, slug, username }) {
   return (dispatch) => {
     dispatch(checkVersion());
+    dispatch(markPending(payload.pokemon, slug, username));
 
     return API.delete(`${Config.API_HOST}/captures`, payload)
       .then(() => dispatch(markCaptured(false, payload.pokemon, slug, username)));
@@ -37,6 +41,10 @@ export function listCaptures ({ slug }, username) {
 
 export function setCaptures (captures, slug, username) {
   return { type: SET_CAPTURES, captures, slug, username };
+}
+
+export function markPending (pokemon, slug, username) {
+  return { type: MARK_PENDING, captured: false, pokemon, slug, username };
 }
 
 export function markCaptured (captured, pokemon, slug, username) {
--- a/app/actions/capture.js
+++ b/app/actions/capture.js
@@ -6,7 +6,6 @@ export const MARK_PENDING = 'MARK_PENDING';
 export const MARK_CAPTURED = 'MARK_CAPTURED';
 export const SET_CAPTURES = 'SET_CAPTURES';
 
-
 export function createCaptures ({ payload, slug, username }) {
   return (dispatch) => {
     dispatch(checkVersion());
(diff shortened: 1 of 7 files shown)


app/styles/tracker.scss
```diff
@@ -483,6 +483,10 @@ header {
   width: $pokemon-box-size;
   height: $pokemon-box-size;
 
+  &.pending {
+    background: linear-gradient(0deg, $brand-primary-light 0%, rgba(0, 0, 0, 0) 100%);
```
ami, line 487: let's use `background: rgba($brand-primary-light, .5);` here please!

app/styles/main.scss
```diff
@@ -843,6 +843,10 @@ select {
   }
 
   .pokemon {
+    &.pending {
+      background: linear-gradient(0deg, $night-mode-primary 0%, rgba(0, 0, 0, 0) 100%);
```
ami, line 847: similarly for night mode, `background: rgba($night-mode-primary, .5);`

ami (changes requested):
thank you for the contribution @Fleker! we don't use gradients on the rest of the site so i suggested minor changes to keep it stylistically similar.

i'll let @robinjoseph08 review the code but other than the colors, functionally seems sound to me.

app/reducers/captures.js
```diff
@@ -15,6 +15,28 @@ export function captures (state = {}, action) {
...
+              ...state[action.username].dexesBySlug[action.slug],
+              captures: state[action.username].dexesBySlug[action.slug].captures.slice()
+            }
+          }
+        }
+      };
+      const pendingIndex = pendingState[action.username].dexesBySlug[action.slug].captures.findIndex((c) => c.pokemon.id === action.pokemon[0]);
+      for (let i = pendingIndex, count = 0; count < action.pokemon.length; i++) {
+        if (action.pokemon.indexOf(pendingState[action.username].dexesBySlug[action.slug].captures[i].pokemon.id) !== -1) {
+          pendingState[action.username].dexesBySlug[action.slug].captures[i].captured = action.captured;
```
robinjoseph08, line 34: instead of passing in `captured: false` in the action, since it's always going to be false, it should probably hard-code `false` here, similarly to how `pending` is being hard-coded to `true`

app/components/pokemon.jsx
```diff
@@ -5,7 +5,7 @@ import { faInfo }                   from '@fortawesome/free-solid-svg-icons';
 import { useDispatch, useSelector } from 'react-redux';
 
 import { ReactGA }                        from '../utils/analytics';
-import { createCaptures, deleteCaptures } from '../actions/capture';
+import { createCaptures, deleteCaptures, markPending } from '../actions/capture';
```
robinjoseph08, line 8: it doesn't look like `markPending` is being used in this file, so let's remove it as an import

app/actions/capture.js
```diff
@@ -39,6 +43,10 @@ export function setCaptures (captures, slug, username) {
   return { type: SET_CAPTURES, captures, slug, username };
 }
 
+export function markPending (pokemon, slug, username) {
+  return { type: MARK_PENDING, captured: false, pokemon, slug, username };
```
robinjoseph08, line 47: once we're hard-coding `false` in the reducer, we can remove `captured: false` from here

app/actions/capture.js
```diff
@@ -2,12 +2,15 @@ import { API }          from '../utils/api';
 import { Config }       from '../../config';
 import { checkVersion } from './utils';
 
+export const MARK_PENDING = 'MARK_PENDING';
 export const MARK_CAPTURED = 'MARK_CAPTURED';
 export const SET_CAPTURES = 'SET_CAPTURES';
 
+
```
robinjoseph08, line 9: looks like a stray extra line here. the linter should've caught this, but it looks like CI wasn't running for forked repos, so it wasn't clear. sorry about that. i've enabled CI runs on forks now, so small things will this will be apparent immediately for you! you can also run `yarn lint` locally to see if it passes as well

robinjoseph08 (commented):
thanks @Fleker! just some small stuff in the code, but overall looks good!

Fleker:
Resolved those comments. Thanks.

robinjoseph08 (approved):
Looks great! Thanks a ton for addressing the changes! We appreciate the contribution!

We'll be deploying it soon. Once it's deployed, I'll comment back on this PR to let you know.

robinjoseph08:
This has been deployed! Thanks again!

Issue #420, opened 2021-10-24
Suggestion: Sort by Pokemon Regions


ThePharaohArt:
I think the list of national dex would benefit greatly if it was possible to sort by pokemon that are from specific regions:
Kanto, Sinnoh, Hoenn, Johto etc.
The pokemon that appeared first in whatever region will be on that regions list, so there would be no dups.

ami:
Hey @ArtexJay, a national dex is already inherently sorted by Region. Could you explain your use case more?

ArtexJay:
> Hey @ArtexJay, a national dex is already inherently sorted by Region. Could you explain your use case more?

it's sorted by every 30 pokemon number in the national pokedex. Rather than having the number 001-030 as the header for a set of pokemon i'd rather have the pokemon groups separated by region.
If i want to view the list of only Kanto region pokemon, i should be able to  or view only Johto, Unova etc

Right now I at least cannot clearly distinguish when one region ends and when a new one begins in the national dex.

Azekthi:
While not intentional, this would also be a step towards #372 since the dex in GO is separated by region.

The only thing with this is that Meltan and Melmetal need to be in their own region named "Unknown" or something similar, since they currently have no origin region.

PR #436, opened 2022-02-05
feat(dex): Sort Pokemon by Generation

Garethp:
Adds a button to sort the Pokemon in a dex by Generation as well as Dex Number. 

Fixes: #420

### DIFF
--- a/app/components/dex.jsx
+++ b/app/components/dex.jsx
@@ -1,18 +1,19 @@
-import PropTypes       from 'prop-types';
-import { Link }        from 'react-router-dom';
-import { useMemo }     from 'react';
-import { useSelector } from 'react-redux';
+import PropTypes             from 'prop-types';
+import { Link }              from 'react-router-dom';
+import { useMemo, useState } from 'react';
+import { useSelector }       from 'react-redux';
 
-import { Box }           from './box';
-import { DonatedFlair }  from './donated-flair';
-import { FriendCode }    from './friend-code';
-import { Header }        from './header';
-import { Notification }  from './notification';
-import { Progress }      from './progress';
-import { ReactGA }       from '../utils/analytics';
-import { Scroll }        from './scroll';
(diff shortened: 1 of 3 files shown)


Garethp:
I've been trying to start my living dex from the beginning and I want each generation of Pokemon to be caught in the generation they were introduced, however this is a bit of a problem since the earliest games supported are Gen 6. I was using the ORAS National Dex for Emerald but it made figuring out where to put which Pokemon in my boxes a bit more difficult since I had to skip over the older Gens and the newer Gens (such as Gen 4 evolutions/babies of Gen 3 Pokemon). This functionality makes it a lot easier to use the newer dex definitions to track for older games.

I think with this all you need is a dex type that returns all Pokemon for all games and you'd have a nice way to track your entire living dex in one place, rather than splitting them up into multiple dexes

ami:
hey @Garethp!

our preferred way to solve this problem is to actually add support for the older games (including capture info), as opposed to adding sorting. we're not trying to optimize for using the current gen dexes to track older games.

thanks anyway for the contribution! we'll be cleaning up our submitted issues so it can be more clear which things are actionable or not.

Garethp:
I still believe that sorting by generation has it's own use-case, especially if you want to have each generation in it's own box (for example, in Home I'd rather have my Gen 2 Pokemon in a separate box from my Gen 3 Pokemon. Being able to see that Treeko is the first Pokemon in it's own box rather than at the end of row 2 of a box that starts with Miltank is useful to me. I personally end up running my forked webapp locally and have it talk to your API for this exact reason.

That being said, I'm aware this is an open source project and for the issue of supporting older games you'd rather add support directly such as adding capture info. How can I contribute that data? It doesn't look as if the full data set is in any of the Github  repos. If possible, I'd like to help contribute to extending this project's support for games and features in ways that you prefer.

Issue #456, opened 2022-05-27
PWA feature to install Pokedex Tracker as App on desktop & mobile


walmeidaw:

leonardoalemax:
hey, can I help with that?

leonardoalemax:
@walmeidaw are you doing it?

walmeidaw:
> @walmeidaw are you still doing it? 
> 
> @Fleker also we will need the right icon images to go with the manifest, if Walmeidaw does not answer, I will send a PR with it. I have done this for my work many times. not a big issue. 

I'm not working with it, was just a suggestion, you can send you PR

Fleker:
Yeah the initial work needed to do this is not much. I imagine all the pieces you need are already in other places.

walmeidaw:
> hey, can I help with that?

If you would do it by four hands call me, I have some XP with PWA :)

Fleker:
Has anyone decided to take this up? I could probably get out a quick proof-of-concept in the next week.

walmeidaw:
> Has anyone decided to take this up? I could probably get out a quick proof-of-concept in the next week.

I think no one is working on it for now, you can do your poc. If need someone to test and help, call me

Fleker:
A quick preview of it running.
![image](https://user-images.githubusercontent.com/3291635/213053818-9103abf2-08db-47f8-a5f0-d3251791b2fc.png)

robinjoseph08:
Hey everyone, sorry for the delay on following up on this. Thanks to the work that @Fleker did, I've now merged and deployed basic PWA functionality for Pokedex Tracker, so you should now be able to install it as an app on your devices! If you find any issues with it, feel free to create a new issue for it.

PR #482, opened 2023-01-19
Add manifest.json file to enable PWA

Fleker:
Addresses #456 by adding a `manifest.json` and a barebones `sw.js` service worker

The PWA jumps by default to your profile page using `/u/me` as a new route that uses the session user details.

Include additional PWA polish such as screenshots and shortcuts.


robinjoseph08:
Hey @Fleker, apologies for not getting to this sooner. I really appreciate you taking the time to work on this.

Recently, I did a major overhaul of this repo to upgrade all the packages, switch to TypeScript, organize the files better, etc., so it would've been rough to rebase it, so I carried your changes over to a separate PR, and I merged and deployed that, so Pokedex Tracker should now be installable as a PWA. I'm not well-versed in PWAs, but it was a lot easier to understand with your PR as a base, so once again, I appreciate the work that you did.

One change I wanted to note was that instead of adding a `/u/me` that rendered your profile, I ended up adding a `/profile` route that just redirected to the correct location. I chose `/profile` over `/u/me` because that route would prevent anyone from viewing the user `me`'s profile which is unfortunate. And the redirect seemed like a cleaner approach so that we could send the user to the login page if they weren't logged in.

The 2 things that I didn't copy over were:
- I left off the `screenshots` property of the manifest. It seems likes just for promotional purposes, not functionality purposes, so I opted to leave it off for now.
- I omitted the service worker changes. It seemed like just having an empty service worker [doesn't provide any added functionality by itself](https://web.dev/learn/pwa/service-workers/#capabilities), and since adding in additional functionality in the service worker seemed like a larger lift, I opted to not do it for now to at least get this shipped faster. If we find that there's additional functionality that would be worthwhile for us to add (e.g. offline capabilities), we'll have separate issues to keep track of those.

Thanks again for putting up this PR, and sorry I didn't get to it sooner!

PR #518, opened 2023-11-17
feat(dex-types): support customizations

robinjoseph08:
(*꒦ິ꒳꒦ີ)

### what

- add support for dex type customizations
- make smol changes to support sv dlc


ami:
lgtm

